PR #9, opened 2022-03-23
l1tol2message getMessageInputs method + update eth deposit tests

DZGoldman:


src/lib/message/L1ToL2Message.ts
```diff
@@ -237,6 +249,53 @@ export class L1ToL2MessageReader extends L1ToL2Message {
     return currentTimestamp.gte(timeoutTimestamp)
   }
 
+  public async getMessageInputs(): Promise<L1toL2MessageInputs> {
```
yahgwai, line 252: Some docs here. Also maybe call it getInputComponents() or getInputs(), since we're already on the message

src/lib/message/L1ToL2Message.ts
```diff
@@ -65,6 +66,17 @@ export enum L1ToL2MessageStatus {
   EXPIRED = 5,
 }
 
+export interface L1toL2MessageInputs {
+  destinationAddress: string
```
yahgwai, line 70: Also some docs on this interface, and each of it's properties

yahgwai (approved):

PR #10, opened 2022-03-23
make L1ToL2Message.status method public

DZGoldman:


yahgwai (approved):
I guess this is needed somewhere externally? I've been considering getting rid of this and just leaving `waitForStatus`

DZGoldman:
`status` is useful! i.e., being able to check it without waiting (to show that it's still, i.e., pending), came up naturally in the UI

PR #11, opened 2022-03-24
Address alias bug

yahgwai:
Our address aliasing didnt overflow properly for this address: 0xFfC98231ef2fd1F77106E10581A1faC14E29d014

### DIFF
--- a/src/lib/utils/lib.ts
+++ b/src/lib/utils/lib.ts
@@ -1,3 +1,4 @@
+import { getAddress as getAddress } from '@ethersproject/address'
 import { utils } from 'ethers'
 import { ADDRESS_ALIAS_OFFSET } from '../dataEntities/constants'
 import { ArbTsError } from '../dataEntities/errors'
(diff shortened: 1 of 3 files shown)


src/lib/utils/lib.ts
```diff
@@ -15,17 +20,39 @@ export const throwIfNotAddress = (address: string) => {
...
+  // we use BigInts in here and apply to allow for proper under/overflow behaviour
+  // BigInt.asUintN calculates the correct positive modulus
+  return getAddress(
+    '0x' +
+      BigInt.asUintN(
+        ADDRESS_BIT_LENGTH,
+        BigInt(l2Address) - ADDRESS_ALIAS_OFFSET_BIG_INT
+      )
+        .toString(16)
+        .padStart(ADDRESS_NIBBLE_LENGTH, '0')
```
fredlacs, line 56: has a bit of repeated code between here and the apply function.
the only difference with them is either the `-` or `+`

could have an intermediate function that is used with both

fredlacs (dismissed):
looking good, just one small bit of code we can DRY up

src/lib/utils/lib.ts
yahgwai: updated

DZGoldman (approved):
lgtm

PR #17, opened 2022-03-31
search log for first successful redeem

gzeoneth:
- [x] Search log for first redeem attempt if gas underpriced
- [x] Optimize eth_getLogs rpc calls with smaller block range
- [x] Add separate function for auto-redeem and successful-redeem


src/lib/dataEntities/constants.ts
```diff
@@ -46,3 +46,6 @@ export const DISABLED_GATEWAY = '0x0000000000000000000000000000000000000001'
  * isArbitrumEnabled which returns this value. Intger: 0xa4b1
  */
 export const CUSTOM_TOKEN_IS_ENABLED = 42161
+
+// https://github.com/OffchainLabs/nitro/blob/e16d52d4702b73d3fd105ba7b570d1d082a4ad74/arbos/retryables/retryable.go#L21
+export const RETRYABLELIFETIMESECONDS = 7 * 24 * 60 * 60
```
DZGoldman, line 51: could store this in the `L2Network`s, since it's technically a per-chain param

src/lib/message/L1ToL2Message.ts
```diff
@@ -245,17 +246,50 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+      let fromBlock = await this.l2Provider.getBlock(fromBlockNumber)
+      let increment: number = 1000
+      const creationBlock = await this.l2Provider.getBlock(creationBlockNumber)
+      const maxBlock = await this.l2Provider.getBlockNumber()
+      while(fromBlockNumber <= maxBlock) {
+        const toBlockNumber = Math.min(fromBlockNumber + increment, maxBlock)
+        redeemEventLogs = await this.l2Provider.getLogs( {fromBlock: fromBlockNumber, toBlock: toBlockNumber, topics: [redeemTopic, ticketId]})
+        if (redeemEventLogs.length != 0) break
+        const toBlock = await this.l2Provider.getBlock(toBlockNumber)
+        if ((toBlock.timestamp - creationBlock.timestamp) > RETRYABLELIFETIMESECONDS) break
```
DZGoldman, line 268: we could theoretically get a false negative if a `keepalive` is used; not sure we need/want to handle this tho

src/lib/message/L1ToL2Message.ts
```diff
@@ -245,17 +246,50 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+        const processedSeconds = (toBlock.timestamp - fromBlock.timestamp)
+        if (processedSeconds != 0) {
+          // find the increment that cover ~ 1 day
+          increment = increment * RETRYABLELIFETIMESECONDS / 7 / processedSeconds
```
DZGoldman, line 272: Sanity check: so iterations of `increment` would go:
1. 1000 blocks
2. ~ # of blocks in 1 day
3. ~ # of blocks in 1 day
4. ... etc.

tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "noImplicitThis": true,
     "resolveJsonModule": true,
     "esModuleInterop": true,
+    "skipLibCheck": true,
```
DZGoldman, line 17: y tho?

DZGoldman (commented):
looks good, only minor questions/ comments

src/lib/message/L1ToL2Message.ts
gzeoneth: can add a optional flag that also look for the `LifetimeExtended` event to adjust the deadline accordingly, don't think we should make it the default behavior as it will double the `eth_getLog` calls

src/lib/message/L1ToL2Message.ts
gzeoneth: unless toBlock.timestamp == fromBlock.timestamp, then it would keep requesting 1000 blocks
I think we can change the `1000` constant if we have an estimate of how much block to expect between a timestamp update

tsconfig.json
gzeoneth: mocha and jest have a type conflict that break my build in https://github.com/OffchainLabs/retryable-tx-panel/pull/12

src/lib/dataEntities/constants.ts
```diff
@@ -46,3 +46,6 @@ export const DISABLED_GATEWAY = '0x0000000000000000000000000000000000000001'
  * isArbitrumEnabled which returns this value. Intger: 0xa4b1
  */
 export const CUSTOM_TOKEN_IS_ENABLED = 42161
+
+// https://github.com/OffchainLabs/nitro/blob/e16d52d4702b73d3fd105ba7b570d1d082a4ad74/arbos/retryables/retryable.go#L21
```
yahgwai, line 50: We can use /* */ style comments to make them available when hovering over. Also should include a humanb readable description

src/lib/message/L1ToL2Message.ts
```diff
@@ -245,17 +246,50 @@ export class L1ToL2MessageReader extends L1ToL2Message {
     const creationReceipt =
       retryableCreationReceipt ||
       (await this.l2Provider.getTransactionReceipt(this.retryableCreationId))
-
-    if (creationReceipt) {
-      const l2Receipt = new L2TransactionReceipt(creationReceipt)
-      const redeemEvents = l2Receipt.getRedeemScheduledEvents()
+    if (creationReceipt){
+      const creationBlockNumber = creationReceipt.blockNumber
```
yahgwai, line 250: I'm not sure what the advantage is of replacing the old code with this new section. Both only fetch the first redeem attempt, but the previous version did it without any log calls at all.

I think what we need is a new function to fetch the first successful redeem, instead of just the first scheduled one. For this you need to find the last redeem that took place, and check if it was successful.

gzeoneth: There are no redeem attempt in nitro if it the auto redeem fail.

yahgwai: A user can later call `redeem` manually on ArbRetryable.sol, which will schedule a new redeem.

gzeoneth: The old code look for the redeem attempt only in the ticket creation tx, so it won't be able to find manual redemption.

yahgwai: Yeah, but I think the new code and old are trying to do different jobs, the old code returns the initial redeem attempt. The new code should be looking for the first successful redeem. So I think they should be separated into different functions. 

I think the new code doesn't quite get the first successful redeem, as it misses successful redeems that happen outside the current block interval being checked. The code does the following:

1. Looks for any redeem attempts in the first 1000 blocks
2. If it finds one (it definitely will since the ticket creation itself causes one redeem attempt), it breaks from the while loop
3. It then returns (but only if there's one redeem attempt: https://github.com/OffchainLabs/arbitrum-sdk/pull/17/files#diff-44f6a12b564108aa7000a7e940d0918990f51cd4bd32feacff3a57d8448728a6R287)

So I think scenarios like the following dont get found:
block 0: ticket is created, redeem scheduled but failed 
block 500: ticket redeemed manually, but redemption failed
block 1001: ticket redeemed manualy, redemption successful.

I think the new code would return the redemptions from block 0 and block 500, then break from the while loop and miss the successful redemption on block 1001

src/lib/message/L1ToL2Message.ts
yahgwai: I think we should look for the last redeem attempt and see if it was successful. If it wasnt then we can see if a keep alive occurred, and if it did continue to look for more redeems. I dont think we should have an optional flag though, since i dont think there are any occasions when a user would not want to know if a redeem happened in an extended period.

tsconfig.json
fredlacs: to fix that you can do `yarn install --production` in the panel instead of skipLibCheck here

src/lib/message/L1ToL2Message.ts
gzeoneth: > If it finds one (it definitely will since the ticket creation itself causes one redeem attempt), it breaks from the while loop
This is incorrect, at least if the redeem attempt failed due to low gas price there will not be redeem attempt.

yahgwai: Yep, good point

gzeoneth: There are 3 case upon retryable creation
1. Auto redeem success
2. Auto redeem failed due to gas parameters
3. Auto redeem failed due to l2 call revert
1 and 3 will emit `RedeemScheduled` that we can easily catch
2 will not emit `RedeemScheduled` and will require a log scan, which this PR solves

We will also need to add a function that return the first successful attempt, correct?

src/lib/message/L1ToL2Message.ts
gzeoneth: Agree but how about we add the keepalive in another PR?

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
 
+  private async getFirstRedeem(
+    retryableCreationReceipt?: TransactionReceipt,
+    includeFail?: boolean
+  ): Promise<TransactionReceipt | null> {
+    includeFail = includeFail || false
+    const l2Network = await getL2Network(this.l2Provider)
+    const creationReceipt =
+      retryableCreationReceipt ||
+      (await this.l2Provider.getTransactionReceipt(this.retryableCreationId))
```
yahgwai, line 294: Instead of potentially looking up again, lets make `getRetryableCreationReceipt` cache the results of trying to fetch the receipt

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+                gasDonor: string
+              }
+          )
+          if(includeFail) {
+            return await this.l2Provider.getTransactionReceipt(redeemEvents[0].retryTxHash)
+          }
+          const successfulRedeem = (await Promise.all(redeemEvents.map(
+            async e => await this.l2Provider.getTransactionReceipt(e.retryTxHash)))).filter(
+              r => r.status === 1)
+          if (successfulRedeem.length > 1) throw new Error("More than 1 successful redeem?")
```
yahgwai, line 331: Should throw an `ArbTsError`. Should probably make the error message more descriptive too:
`Unexpected number of successful redeems. Expected only one redeem for transaction ${transactionHash}, but found ${successfulRedeem.length}.`

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+    if (creationReceipt){
+      const creationBlockNumber = creationReceipt.blockNumber
+      const ticketId = creationReceipt.transactionHash
+      const iFace = new Interface([
+        'event RedeemScheduled(     bytes32 indexed ticketId,     bytes32 indexed retryTxHash,     uint64 indexed sequenceNum,     uint64 donatedGas,     address gasDonor )',
+      ])
+      const redeemTopic = iFace.getEventTopic('RedeemScheduled')
+
+      let redeemEventLogs: ethers.providers.Log[] = []
+      let fromBlockNumber: number = creationBlockNumber
```
yahgwai, line 306: I dont think we need the `fromBlockNumber` variable, could just do `fromBlock.number` where we need that.

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+              r => r.status === 1)
+          if (successfulRedeem.length > 1) throw new Error("More than 1 successful redeem?")
+          if (successfulRedeem.length == 1) return successfulRedeem[0]
+        }
+        const toBlock = await this.l2Provider.getBlock(toBlockNumber)
+        if ((toBlock.timestamp - creationBlock.timestamp) > l2Network.retryableLifetimeSeconds) break
+        const processedSeconds = (toBlock.timestamp - fromBlock.timestamp)
+        if (processedSeconds != 0) {
+          // find the increment that cover ~ 1 day
+          increment = increment * l2Network.retryableLifetimeSeconds / 7 / processedSeconds
```
yahgwai, line 339: Need to do a Math.floor/ceil here to keep `increment` as an integer

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+            async e => await this.l2Provider.getTransactionReceipt(e.retryTxHash)))).filter(
+              r => r.status === 1)
+          if (successfulRedeem.length > 1) throw new Error("More than 1 successful redeem?")
+          if (successfulRedeem.length == 1) return successfulRedeem[0]
+        }
+        const toBlock = await this.l2Provider.getBlock(toBlockNumber)
+        if ((toBlock.timestamp - creationBlock.timestamp) > l2Network.retryableLifetimeSeconds) break
+        const processedSeconds = (toBlock.timestamp - fromBlock.timestamp)
+        if (processedSeconds != 0) {
+          // find the increment that cover ~ 1 day
```
yahgwai, line 338: I guess we're not finding a 1 day increment here, but rather the `l2Network.retryableLifetimeSeconds / 7`. Whilst i guess for our current networks this is 1 day, it seems rather arbitrary when we put it like that. If we really only want to search 1 day wouldn't it be better to do:
```
increment *= Math.ceil(86400 / processedSeconds)
```

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+      let redeemEventLogs: ethers.providers.Log[] = []
+      let fromBlockNumber: number = creationBlockNumber
+      let fromBlock = await this.l2Provider.getBlock(fromBlockNumber)
+      let increment: number = 1000
+      const creationBlock = await this.l2Provider.getBlock(creationBlockNumber)
+      const maxBlock = await this.l2Provider.getBlockNumber()
+      while(fromBlockNumber <= maxBlock) {
+        const toBlockNumber = Math.min(fromBlockNumber + increment, maxBlock)
+        redeemEventLogs = await this.l2Provider.getLogs( {fromBlock: fromBlockNumber, toBlock: toBlockNumber, topics: [redeemTopic, ticketId]})
+        if (redeemEventLogs.length != 0) {
```
yahgwai, line 314: I think we dont need this `if` statemate since map, and filter, will just be a noop for an empty array

src/lib/message/L1ToL2Message.ts
```diff
@@ -228,18 +230,38 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+    return await this.getFirstRedeem(undefined, true)
+  }
+
+    /**
+   * Receipt for the successful l2 transaction created by this message. See L1ToL2Message.l2TxHash
+   * May throw an error if the l2 transaction has yet to be executed, which is the case if
+   * the retryable ticket has not been created and redeemed.
+   * @returns TransactionReceipt of the first successful redeem if exists, otherwise null
+   */
+  public async getSuccessfulRedeemAttempt(): Promise<TransactionReceipt | null> {
```
yahgwai, line 248: Maybe we can discuss later, but I think this is probably the only public function we need (although we should rename it to `getSuccessfulRedeem` since it only returns successful redeems). We may also want to keep `getAutoRedeem` if for some reason callers want to do a quick check of the auto redeem but are cautious about doing a heavy query, which the success case would entail.

src/lib/dataEntities/networks.ts
```diff
@@ -178,6 +179,7 @@ export const l2Networks: L2Networks = {
     confirmPeriodBlocks: 45818,
     rpcURL: process.env['ARB_ONE_RPC'] || 'https://arb1.arbitrum.io/rpc',
     isCustom: false,
+    retryableLifetimeSeconds: 7 * 24 * 60 * 60
```
fredlacs, line 182: could create a `const SEVEN_DAYS_IN_SECONDS = 7 * 24 * 60 * 60` to make this clearer

src/lib/message/L1ToL2Message.ts
fredlacs: yea agreed that it can be useful to have a `getAutoRedeem`

src/lib/message/L1ToL2Message.ts
```diff
@@ -259,6 +283,68 @@ export class L1ToL2MessageReader extends L1ToL2Message {
     return null
   }
 
+  private async getFirstRedeem(
+    retryableCreationReceipt?: TransactionReceipt,
+    includeFail?: boolean
```
fredlacs, line 288: If we do keep this, we can give it a default value
```suggestion
    includeFail = false
```

instead of
```
    includeFail = includeFail || false
```

src/lib/message/L1ToL2Message.ts
fredlacs: yea agreed this will prob be cleaner

src/lib/message/L1ToL2Message.ts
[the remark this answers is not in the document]
fredlacs: I think the reason for the "includeFail" is for the `getFirstRedeemAttempt` to use this same logic here

src/lib/message/L1ToL2Message.ts
fredlacs: yea we should probably use BigNumber instead of js's number

src/lib/message/L1ToL2Message.ts
gzeoneth: Agree that `getSuccessfulRedeemAttempt` should be the default, but for example retryable-dashboard would want `getFirstRedeemAttempt` in case there are no successful redeem so we can look at the revert reason. `getAutoRedeemAttempt` might be used to debug auto-redeem issue.

src/lib/message/L1ToL2Message.ts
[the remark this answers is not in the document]
gzeoneth: yes that's the reason

src/lib/message/L1ToL2Message.ts
gzeoneth: Basically `getFirstRedeemAttempt`

gzeoneth: `getFirstRedeemAttempt` fallback to `getAutoRedeemAttempt` if there are auto-redeem attempt

src/lib/message/L1ToL2Message.ts
yahgwai: Ah yeh I see. If you wanted you could detect that by checking that fromBlock.number is the creationBlockNumber eg:
```
await this.l2Provider.getLogs( {fromBlock: fromBlock.number === creationBlockNumber ? fromBlock.number : fromBlock.number + 1, toBlock: toBlockNumber, topics: [redeemTopic, ticketId]})
```

gzeoneth: I think using a separate variable looks cleaner, no?

yahgwai: Yeh, could be a clearer, but I think both ways are a bit messy for different reasons. 

For the separate var I was more concerned about maintainability, having two variables feels like it's potentially bug prone - ie if fromBlock and fromBlockNumber get out of sync with each other for some reason.

src/lib/message/L1ToL2Message.ts
gzeoneth: > But why is the first manual redeem attempt interesting, compared to for example the last one?

There are no guarantee `getSuccessfulRedeemAttempt` will return the last one, it may as well return nothing if there are no successful redeem attempt.

Say someone sent a L1 -> L2 tx that is both underpriced and may revert in L2. Since the auto-redeem fail, `getAutoRedeemAttempt` and `getSuccessfulRedeemAttempt` return null. If the user attempt to manual redeem, and the l2 tx reverted, he will need to use `getFirstRedeemAttempt` to find the reverting l2 tx.

gzeoneth: But if what you mean is to have `getSuccessfulRedeemAttempt` return the last redeem attempt (which the success attempt must be the last), we can do that too.
How about
`getSuccessfulRedeemAttempt` -> `getLastRedeem`
keep `getAutoRedeemAttempt` for debug 
drop `getFirstRedeemAttempt`  ?

gzeoneth: My concern is that `getLastRedeem` will force us to scan through all the block within `keepalive` for any failed retryable, but assuming the redeem is going to fail for similar reason, `getFirstRedeemAttempt` should be enough for one to debug what's wrong

yahgwai: I'm ok with leaving `getSuccessfulRedeem` and just dropping the `getFirstRedeemAttempt`. We'll lose the ability to get the first manual redeem attempt (in the case that there wasnt an auto redeem), but my argument is that that's not really an interesting special case.

If we do want to provide info about the failing manual redeems, then I would suggest we focus on the last redeem, or provide access to all failed redeems. I think we can do that stuff in a later pr, if we choose to do it all.

gzeoneth: ok resolved in cad0fb7ac

src/lib/message/L1ToL2Message.ts
gzeoneth: resolved in 6d914f3a4
turns out we can use `fromBlock.number + 1` for both case as `fromBlock.number` is covered by getAutoRedeem shortcut

gzeoneth:
resolved suggestions and ready for review

src/lib/message/L1ToL2Message.ts
```diff
@@ -222,24 +225,44 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
-  public async getFirstRedeemAttempt(): Promise<TransactionReceipt | null> {
-    return await this.getFirstRedeem()
+  public async getAutoRedeemAttempt(): Promise<TransactionReceipt | null> {
+    return await this.getAutoRedeem()
   }
 
   // CHRIS: TODO: read ALL the comments in arbitrum sdk and check for accuracy with nitro
 
-  private async getFirstRedeem(
+  private async getAutoRedeem(
```
yahgwai, line 265: I think we can get rid of the `retryableCreationReceipt?: TransactionReceipt` arg (since we can use the internal cached method). Then the public and private functions will have the same signature, so I think can just merge them / get rid of the private variant

src/lib/message/L1ToL2Message.ts
```diff
@@ -222,24 +225,44 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
-   * Receipt for the l2 transaction created by this message. See L1ToL2Message.l2TxHash
+    /**
+   * Receipt for the successful l2 transaction created by this message. See L1ToL2Message.l2TxHash
    * May throw an error if the l2 transaction has yet to be executed, which is the case if
    * the retryable ticket has not been created and redeemed.
-   * @returns
+   * @returns TransactionReceipt of the first successful redeem if exists, otherwise null
+   */
+  public async getSuccessfulRedeem(): Promise<TransactionReceipt | null> {
+    return await this.getFirstSuccessfulRedeem()
```
yahgwai, line 250: As with the autoRedeem, I think we can move the functionality from the priate method here and delete the private method.

src/lib/message/L1ToL2Message.ts
```diff
@@ -222,24 +225,44 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
    * @returns
    */
-  public getRetryableCreationReceipt(): Promise<TransactionReceipt> {
-    return this.l2Provider.getTransactionReceipt(this.retryableCreationId)
+  public async getRetryableCreationReceipt(    
+    confirmations?: number,
+    timeout?: number
+  ): Promise<TransactionReceipt> {
+    if (!this.retryableCreationReceipt){
+      if (timeout){
```
yahgwai, line 233: Should we do `(timeout || confirmations)` here?

yahgwai (approved):
Nice. Added some minor comments, but apart from that looks good!

gzeoneth:
Thanks @yahgwai, pushed a commit to fix the recommendations

PR #21, opened 2022-04-05
Add node deadline estimation for L2 to L1 messages

fredlacs:
Added a utility method to calculate node deadline following the implementation from the [token bridge]( https://github.com/OffchainLabs/arb-token-bridge/blob/02817aff63b87f7a6dabe5fa4f6457fc024460a9/packages/token-bridge-sdk/src/hooks/useArbTokenBridge.ts#L1052) and substituting [subgraphs](https://github.com/OffchainLabs/arbitrum-subgraphs/blob/e6be72526329c5a8e04b39393a258e39c5c49363/packages/arb-bridge-eth/src/mapping.ts#L200-L210) for event queries

### DIFF
--- a/src/lib/message/L2ToL1Message.ts
+++ b/src/lib/message/L2ToL1Message.ts
@@ -336,52 +336,46 @@ export class L2ToL1MessageReader extends L2ToL1Message {
       this.l1Provider
     )
 
-    try {
-      const proof = await this.tryGetProof(l2Provider)
-      // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time.
-      if (proof === null) return BigNumber.from(network.confirmPeriodBlocks)
-      // we can't check if the L2 to L1 tx isSpent on the outbox, so we instead try executing it
-      if (await this.hasExecuted(proof)) return BigNumber.from(0)
-      // 60seconds * 60 minutes * 24hrs * 8 days
-      const EIGHT_DAYS_IN_BLOCKS =
-        (60 * 60 * 24 * 8) / (await getL1Network(this.l1Provider)).blockTime
-      const latestBlock = await this.l1Provider.getBlockNumber()
+    const proof = await this.tryGetProof(l2Provider)
+    // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time.
+    if (proof === null) return BigNumber.from(network.confirmPeriodBlocks)
+    // we can't check if the L2 to L1 tx isSpent on the outbox, so we instead try executing it
+    if (await this.hasExecuted(proof)) return BigNumber.from(0)
+    // 60seconds * 60 minutes * 24hrs * 8 days
+    const EIGHT_DAYS_IN_BLOCKS =
+      (60 * 60 * 24 * 8) / (await getL1Network(this.l1Provider)).blockTime
+    const latestBlock = await this.l1Provider.getBlockNumber()
 
-      const eventFetcher = new EventFetcher(this.l1Provider)
+    const eventFetcher = new EventFetcher(this.l1Provider)
 
-      const events = (
-        await eventFetcher.getEvents(
-          network.ethBridge.rollup,
-          RollupUserFacet__factory,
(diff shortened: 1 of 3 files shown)


src/lib/message/L2ToL1Message.ts
```diff
@@ -312,6 +315,71 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+      const eventFetcher = new EventFetcher(this.l1Provider)
+
+      const events = (
+        await eventFetcher.getEvents(
+          network.ethBridge.rollup,
+          RollupUserFacet__factory,
+          t => t.filters.NodeCreated(),
+          {
+            fromBlock: latestBlock - EIGHT_DAYS_IN_BLOCKS,
+            toBlock: latestBlock,
```
fredlacs, line 357: could also separate this into a 10 different queries, each being 4000 block range. dispatch in parallel etc

src/lib/message/L2ToL1Message.ts
```diff
@@ -312,6 +315,71 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+    const rollup = RollupUserFacet__factory.connect(
+      network.ethBridge.rollup,
+      this.l1Provider
+    )
+
+    try {
+      const proof = await this.tryGetProof(l2Provider)
+      // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time.
+      // we could check that on the constructor that the L2 to L1 msg is valid
+      if (proof === null) return maxTime
```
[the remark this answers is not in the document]
fredlacs: damn tru

src/lib/message/L2ToL1Message.ts
```diff
@@ -312,6 +315,71 @@ export class L2ToL1MessageReader extends L2ToL1Message {
       await this.waitUntilOutboxEntryCreated(retryDelay)
     }
   }
+
+  // TODO: create version that queries multiple L2 to L1 txs, so a single multicall can make all requests
+  public async getExpectedNodeDeadline(
```
yahgwai, line 320: Maybe we could have a more general name for this? Maybe `getFirstExecutableBlock` or something like that? And include in the comment that this is the deadline of the node which contains this message.

src/lib/message/L2ToL1Message.ts
```diff
@@ -312,6 +315,71 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+          {
+            fromBlock: latestBlock - EIGHT_DAYS_IN_BLOCKS,
+            toBlock: latestBlock,
+          }
+        )
+      )
+        .map(e => e.event)
+        .filter(e => {
+          const afterSendCount = e.assertionBytes32Fields[1][2]
+          return BigNumber.from(afterSendCount).gte(this.batchNumber)
```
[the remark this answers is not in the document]
fredlacs: You're right, expected `assertionIntFields` here.
As for the latter, this is the value we're using for classic on the subgraph that the UI is quering - might change with nitro 
https://github.com/OffchainLabs/arbitrum-subgraphs/blob/e6be72526329c5a8e04b39393a258e39c5c49363/packages/arb-bridge-eth/src/mapping.ts#L210

src/lib/message/L2ToL1Message.ts
fredlacs: yup good point, then we should probably add a bit of padding to the return value to account for the time difference between a node being confirmable, and a validator actually confirming it

src/lib/message/L2ToL1Message.ts
```diff
@@ -312,6 +315,77 @@ export class L2ToL1MessageReader extends L2ToL1Message {
       await this.waitUntilOutboxEntryCreated(retryDelay)
     }
   }
+
+  /**
+   * Estimates the L1 block number in which a particular L2 to L1 tx will be
```
yahgwai, line 320: Can replace `a particular` with `this`

yahgwai (approved):

PR #27, opened 2022-04-08
Add initial impl of nitro node deadline

fredlacs:


src/lib/message/L2ToL1Message.ts
```diff
@@ -352,75 +358,85 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+      l2Network.ethBridge.rollup,
+      [
+        'function getNode(uint64 nodeNum) public view override returns (tuple(bytes32 stateHash, bytes32 challengeHash, bytes32 confirmData, uint64 prevNum, uint64 deadlineBlock, uint64 noChildConfirmedBeforeBlock, uint64 stakerCount, uint64 childStakerCount, uint64 firstChildBlock, uint64 latestChildNumber, uint64 createdAtBlock, bytes32 nodeHash) node)',
+        'event NodeConfirmed(uint64 indexed nodeNum, bytes32 blockHash, bytes32 sendRoot)',
+      ],
       this.l1Provider
     )
 
-    const proof = await this.tryGetProof(l2Provider)
+    const proof = await this.getOutboxProof(l2Provider)
```
yahgwai, line 374: I think this will throw if the node is not confirmed

src/lib/message/L2ToL1Message.ts
```diff
@@ -352,75 +358,85 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
     // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time.
     if (proof === null)
-      return BigNumber.from(network.confirmPeriodBlocks)
+      return BigNumber.from(l2Network.confirmPeriodBlocks)
         .add(ASSERTION_CREATED_PADDING)
         .add(ASSERTION_CONFIRMED_PADDING)
     // we can't check if the L2 to L1 tx isSpent on the outbox, so we instead try executing it
-    if (await this.hasExecuted(proof)) return BigNumber.from(0)
-    const latestBlock = await this.l1Provider.getBlockNumber()
+    if (await this.hasExecuted()) return BigNumber.from(0)
```
yahgwai, line 381: return null instead of 0

src/lib/message/L2ToL1Message.ts
```diff
@@ -352,75 +358,85 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
-    )
-      .map(e => e.event)
-      .filter(e => {
-        const afterSendCount = e.assertionIntFields[1][2]
-        return BigNumber.from(afterSendCount).gte(this.batchNumber)
+          0
+        ),
+        toBlock: 'latest',
+        topics: rollup.interface.encodeFilterTopics(
+          rollup.interface.getEvent('NodeConfirmed'),
```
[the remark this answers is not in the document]
fredlacs: ye I typoed

src/lib/message/L2ToL1Message.ts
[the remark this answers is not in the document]
fredlacs: the thing is that `updateSendRoot` used in the `status` function is expensive. The value won't be cached many times you might be querying for the deadline.
I don't think its worth using the `status` function here, but instead just check if it was executed or not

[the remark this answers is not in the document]
fredlacs: Hmm we might need to do that lookup to infer if the block is available for execution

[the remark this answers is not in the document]
fredlacs: ah looking at this again and `getOutboxProof` already does a `updateSendRoot` query

src/lib/message/L2ToL1Message.ts
fredlacs: removed since we don't actually need that

src/lib/message/L2ToL1Message.ts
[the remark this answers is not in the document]
fredlacs: optimised the query in https://github.com/OffchainLabs/arbitrum-sdk/pull/27/commits/4f973afa92d2afad2b573476113b35aadbf6883a

src/lib/message/L2ToL1Message.ts
```diff
@@ -345,80 +327,88 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+        found = true
+      } else {
+        // TODO: optimise with a binary search
+        logIndex++
+      }
+    }
+
+    // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time
+    // since there isn't a pending node that includes this message yet
+    if (!found)
```
gzeoneth, line 403: unreachable given `while (!found)` above

src/lib/message/L2ToL1Message.ts
```diff
@@ -345,80 +327,88 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
-        .add(ASSERTION_CREATED_PADDING)
-        .add(ASSERTION_CONFIRMED_PADDING)
-    // we can't check if the L2 to L1 tx isSpent on the outbox, so we instead try executing it
-    if (await this.hasExecuted(proof)) return BigNumber.from(0)
+    const status = await this.status(l2Provider)
+    if (status === L2ToL1MessageStatus.EXECUTED) return null
+    if (status === L2ToL1MessageStatus.CONFIRMED) {
+      const latestConfirmed = await rollup.callStatic.latestConfirmed()
+      const node = await rollup.getNode(latestConfirmed)
+      return node.deadlineBlock
```
gzeoneth, line 349: This is probably wrong as the message can be confirmed in earlier node. We may as well just return the latest L1 block here.

src/lib/message/L2ToL1Message.ts
```diff
@@ -345,80 +327,88 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
-      })
-
-    // a node that covers this tx still has not been created
-    if (events.length === 0)
-      return BigNumber.from(network.confirmPeriodBlocks)
+    ).sort((a, b) => a.event.nodeNum.toNumber() - b.event.nodeNum.toNumber())
+
+    let found = false
+    let logIndex = 0
+    while (!found) {
```
gzeoneth, line 382: Should use a for loop over `logs` instead

gzeoneth (approved):
LGTM, added few suggestions
also I think we should retrieve `confirmPeriodBlocks` from rollup contract in `getL2Network`

src/lib/message/L2ToL1Message.ts
fredlacs: my thought here is that the latestConfirmed block's deadline will be <= the current latest L1 block here
the reason i return the deadline instead of latest L1 is because this way it handles better potential lack of sync between different nodes you might be querying.
whoever is consuming the sdk will know the latest block number, and can check if the deadline returned here is <= current block in order to execute

src/lib/message/L2ToL1Message.ts
fredlacs: true, will add a break if index > length

src/lib/message/L2ToL1Message.ts
fredlacs: made it a while loop to make it into a binary search in the future to optimise this

src/lib/message/L2ToL1Message.ts
fredlacs: made it to return `null` instead since this method should only be called if asssertion is pending

src/lib/message/L2ToL1Message.ts
```diff
@@ -376,82 +365,88 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
 
   /**
-   * Estimates the L1 block number in which this L2 to L1 tx will be available for execution
+   * Estimates the L1 block number in which this L2 to L1 tx will be available for execution.
+   * If the message can or already has been executed, this returns null
    * @param l2Provider
-   * @returns expected L1 block number where the L2 to L1 message will be executable
+   * @returns expected L1 block number where the L2 to L1 message will be executable. Returns null if the message can or already has been executed
    */
   public async getFirstExecutableBlock(
```
DZGoldman, line 373: maybe this should rly have estimate in the name? (don't rly care)

src/lib/message/L2ToL1Message.ts
```diff
@@ -376,82 +365,88 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
-    // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time.
-    if (proof === null)
-      return BigNumber.from(network.confirmPeriodBlocks)
-        .add(ASSERTION_CREATED_PADDING)
-        .add(ASSERTION_CONFIRMED_PADDING)
-    // we can't check if the L2 to L1 tx isSpent on the outbox, so we instead try executing it
-    if (await this.hasExecuted(proof)) return BigNumber.from(0)
+    const status = await this.status(l2Provider)
+    if (status === L2ToL1MessageStatus.EXECUTED) return null
+    if (status === L2ToL1MessageStatus.CONFIRMED) return null
```
DZGoldman, line 387: since its getFirstExecutableBlock, maybe CONFIRMED case shouldn't return null? (i.e., confirmed block + confirm padding instead?)

src/lib/message/L2ToL1Message.ts
```diff
@@ -376,82 +365,88 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+    // consistency check in case we change the enum in the future
+    if (status !== L2ToL1MessageStatus.UNCONFIRMED)
+      throw new ArbTsError('L2ToL1Msg expected to be unconfirmed')
+
     const latestBlock = await this.l1Provider.getBlockNumber()
 
     const eventFetcher = new EventFetcher(this.l1Provider)
 
-    const events = (
+    const logs = (
```
DZGoldman, line 399: thinking a logs.length > 0 sanity check (to throw a nice error in the Bad Situation where nodes haven't been created for a long while)

src/lib/message/L2ToL1Message.ts
```diff
@@ -301,54 +300,44 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+      // CHRIS: TODO: handle this case
+      console.log(l2Block['sendRoot'], parsedLog.afterState.sendRoot)
+      throw new ArbTsError("L2 block send root doesn't match parsed log")
     }
 
     const sendRootSize = BigNumber.from(l2Block['sendCount'])
     if (sendRootSize.gt(this.event.position)) {
       this.sendRootSize = sendRootSize
-      this.sendRootHash = parsedLog.sendRoot
+      this.sendRootHash = parsedLog.afterState.sendRoot
```
DZGoldman, line 340: Not really in the scope of this PR, but any not to have updateSendRoot just be getSendRoot (that returns the two values instead of storing them as class variables?)

DZGoldman (approved):
left some comments/suggestions, but none vital

src/lib/message/L2ToL1Message.ts
fredlacs: added in the case

src/lib/message/L2ToL1Message.ts
fredlacs: yea true, `estimateExecutableBlock`

src/lib/message/L2ToL1Message.ts
```diff
@@ -301,54 +300,44 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
-        'event NodeConfirmed(uint64 indexed nodeNum, bytes32 blockHash, bytes32 sendRoot)',
-      ],
-      l1Provider
+      this.l1Provider
     )
 
     // CHRIS: TODO: could confirm in between these calls
-    const latestConfirmedNode = await rollup['latestConfirmed']()
-    const currentBlock = await l1Provider.getBlockNumber()
+    const latestConfirmedNodeNum = await rollup.callStatic.latestConfirmed()
```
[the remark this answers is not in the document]
fredlacs: seen some annoying issues in the past where read-only calls get a `estimateGas` run before if not explicitly "callStatic"

PR #28, opened 2022-04-08
Add nitro typechain gen

fredlacs:
This adds typechain generation for nitro contracts (rollup and precompiles)


gzeoneth (approved):
LGTM

fredlacs:
The current approach is bugged, not building since it can't resolve the OZ in the nitro package which is weird (was working before)

We might need to change the approach here to instead push the compiled contracts to npm

PR #29, opened 2022-04-11
Return as class after L2 to L1 event query

fredlacs:
This adds an optional parameter that allows you to return the L2 to L1 event as a class instance instead of the raw event.
Not sure if this would be the ideal API but from chatting with @spsjvc seems like something along these lines would be useful


DZGoldman:
Nomenclature feels a little weird; what about two separate methods,, getLogs and getEvents, maybe?

PR #32, opened 2022-04-20
writer-from-reader instantiation methods for L1ToL2Message and L2ToL1Message

DZGoldman:


yahgwai (commented):
I'm wondering what the use case of these methods is, since they're quite thin. Isn't it already quite easy to upgrade from reader to writer by doing:
```
const writer = L2ToL1Message.fromBatchNumber(
      l2Signer,
      reader.outboxAddress,
      reader.batchNumber,
      reader.indexInBatch)
```

However, if we do want to make this really easy then these should probably be an instance methods on the readers rather than static:
```
public toWriter(
    l1Signer: Signer
  ): L2ToL1MessageWriter {
    return new L2ToL1MessageWriter(
      l1Signer,
      l2ToL2MessageReader.outboxAddress,
      l2ToL2MessageReader.batchNumber,
      l2ToL2MessageReader.indexInBatch
    )
  }
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@arbitrum/sdk",
-  "version": "1.1.2",
+  "version": "1.1.3",
```
fredlacs, line 3:
```suggestion
  "version": "1.1.4",
```

fredlacs (changes requested):
outdated sdk version here

PR #37, opened 2022-05-01
fix: default l2 gas buffer

gzeoneth:
Temporary workaround of incorrect gas estimation. Also see https://github.com/OffchainLabs/arb-token-bridge/pull/303

### DIFF
--- a/src/lib/message/L1ToL2MessageGasEstimator.ts
+++ b/src/lib/message/L1ToL2MessageGasEstimator.ts
@@ -11,6 +11,7 @@ import { constants } from 'ethers'
 import { utils } from 'ethers'
 
 const DEFAULT_SUBMISSION_PRICE_PERCENT_INCREASE = BigNumber.from(340)
+const DEFAULT_MAX_GAS_PERCENT_INCREASE = BigNumber.from(50)
 
 /**
  * An optional big number percentage increase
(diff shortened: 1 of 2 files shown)


src/lib/message/L1ToL2MessageGasEstimator.ts
```diff
@@ -11,6 +11,7 @@ import { constants } from 'ethers'
 import { utils } from 'ethers'
 
 const DEFAULT_SUBMISSION_PRICE_PERCENT_INCREASE = BigNumber.from(340)
+const DEFAULT_MAX_GAS_PERCENT_INCREASE = BigNumber.from(50)
```
yahgwai, line 14: Lets a comment as to why this is necessary, so that people know when/if they can remove it.

yahgwai (approved):

src/lib/message/L1ToL2MessageGasEstimator.ts
gzeoneth:
```suggestion
// Temporary workaround for incorrect gas estimation from NodeInterface when there is gas refund
const DEFAULT_MAX_GAS_PERCENT_INCREASE = BigNumber.from(50)
```

gzeoneth:
@yahgwai I have added the comment, will let you to merge and publish.

PR #38, opened 2022-05-03
refactor: estimate submission cost without hardcoding value

fredlacs:


src/lib/assetBridger/ethBridger.ts
```diff
@@ -115,22 +111,13 @@ export class EthBridger extends AssetBridger<
     await this.checkL1Network(params.l1Signer)
     await this.checkL2Network(params.l2Provider)
 
-    const gasEstimator = new L1ToL2MessageGasEstimator(params.l2Provider)
```
yahgwai, line 118: Can the l2Provider now be removed from the params?

fredlacs: True, removed

yahgwai (approved):

yahgwai (approved):

PR #40, opened 2022-05-04
feat: update tests and port changes from latest nitro

yahgwai:


integration_test/eth.test.ts
```diff
@@ -41,50 +43,29 @@ describe('Ether', async () => {
...
-      })
-    ).wait()
-
-    const balanceAfter = await l2Signer.provider!.getBalance(
-      await l2Signer.getAddress()
-    )
     const randomBalanceAfter = await l2Signer.provider!.getBalance(
       randomAddress
     )
     expect(randomBalanceAfter.toString(), 'random address balance after').to.eq(
```
fredlacs, line 54: are you sure this is right?
I expected the balance to be `0.001 eth`

src/lib/message/L2ToL1Message.ts
```diff
@@ -72,32 +72,28 @@ export type L2ToL1MessageReaderOrWriter<T extends SignerOrProvider> =
 export class L2ToL1Message {
   public static fromEvent<T extends SignerOrProvider>(
     l1SignerOrProvider: T,
-    event?: L2ToL1TransactionEvent['args'],
-    outboxAddress?: string,
-    batchNumber?: BigNumber,
-    indexInBatch?: BigNumber
+    event: L2ToL1TransactionEvent['args'],
+    outboxAddress: string,
+    isNitro: boolean
```
fredlacs, line 77: instead of having this be user provided, can't we just look at `event.arbBlockNum` and check if its nitro?

src/lib/message/L2ToL1Message.ts
```diff
@@ -72,32 +72,28 @@ export type L2ToL1MessageReaderOrWriter<T extends SignerOrProvider> =
 export class L2ToL1Message {
   public static fromEvent<T extends SignerOrProvider>(
     l1SignerOrProvider: T,
-    event?: L2ToL1TransactionEvent['args'],
-    outboxAddress?: string,
-    batchNumber?: BigNumber,
-    indexInBatch?: BigNumber
+    event: L2ToL1TransactionEvent['args'],
+    outboxAddress: string,
```
fredlacs, line 76: can't we infer the outbox address instead of having it user provided?
if its nitro, its the only outbox
if its pre-nitro, check the batchNum to get the outbox

fredlacs (changes requested):

fork_test/inbox.test.ts
```diff
@@ -1,256 +1,256 @@
...
+//     tx.gasLimit,
+//     tx.maxFeePerGas,
+//     tx.nonce,
+//     tx.to,
+//     tx.value || BigNumber.from(0),
+//     tx.data || '0x'
+//   )
+// }
+
+// describe('Inbox tools', () => {
```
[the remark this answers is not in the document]
yahgwai: Skipping now

src/lib/message/L2ToL1Message.ts
yahgwai: we'd need to make a network call to find that arb block, and this is a constructor so it's only sync.

src/lib/message/L2ToL1Message.ts
yahgwai: We dont have a network object here to lookup the outbox address on

src/lib/message/L2ToL1Message.ts
yahgwai: Alternatively we could change the signature of `public static fromEvent` to make it async

fredlacs: ye we can make fromEvent async and grab the info

src/lib/message/L2ToL1Message.ts
```diff
@@ -66,35 +65,36 @@ export enum L2ToL1MessageStatus {
 export type L2ToL1MessageReaderOrWriter<T extends SignerOrProvider> =
   T extends Provider ? L2ToL1MessageReader : L2ToL1MessageWriter
 
+export type L2ToL1TransactionEvent = {
```
fredlacs, line 68:
```suggestion
import { L2ToL1TransactionEvent as ClassicL2ToL1TransactionEvent } from '@arbitrum/sdk-classic/dist/lib/abi/ArbSys'
```

src/lib/message/L2ToL1Message.ts
```diff
@@ -66,35 +65,36 @@ export enum L2ToL1MessageStatus {
...
  * Base functionality for L2->L1 messages
  */
 export class L2ToL1Message {
   public static fromEvent<T extends SignerOrProvider>(
     l1SignerOrProvider: T,
-    event: L2ToL1TransactionEvent['args'],
-    outboxAddress: string,
-    isNitro: boolean
+    event: L2ToL1TransactionEvent,
+    outboxAddress: string
```
fredlacs, line 88: I think its cleaner if we get the l2SignerOrProvider and use that to get the outboxAddress from the networks object
Users already setup the networks object with that information, so it would be cool if we didn't require it again in public functions

src/lib/message/L2ToL1Message.ts
```diff
@@ -66,35 +65,36 @@ export enum L2ToL1MessageStatus {
...
 /**
  * Base functionality for L2->L1 messages
  */
 export class L2ToL1Message {
   public static fromEvent<T extends SignerOrProvider>(
     l1SignerOrProvider: T,
-    event: L2ToL1TransactionEvent['args'],
-    outboxAddress: string,
-    isNitro: boolean
+    event: L2ToL1TransactionEvent,
```
fredlacs, line 87: we can accept here either the classic or nitro event, I think its cleaner to use the union of those types here
```suggestion
    event: ClassicL2ToL1TransactionEvent['args'] | NitroL2ToL1TransactionEvent['args'],
```

fredlacs (changes requested):

src/lib/message/L2ToL1Message.ts
fredlacs: same as on classic sdk, so lets keep for now

src/lib/message/L2ToL1Message.ts
yahgwai: updated

src/lib/message/L2ToL1Message.ts
```diff
@@ -65,104 +66,62 @@ export enum L2ToL1MessageStatus {
...
       ? new L2ToL1MessageWriter(l1SignerOrProvider, event, outboxAddress)
       : new L2ToL1MessageReader(l1SignerOrProvider, event, outboxAddress)
   }
 
-  public static async getL2ToL1MessageLogs(
+  public static async getEventLogs(
     l2Provider: Provider,
     filter: { fromBlock: BlockTag; toBlock: BlockTag },
-    batchNumber?: BigNumber,
+    positionOrBatchNumber?: BigNumber,
```
fredlacs, line 102: I think its better if we split this into 2 separate parameters

fredlacs: would be cool if you could add some typedoc here explaining when each parameter is expected

src/lib/message/L2Transaction.ts
```diff
@@ -92,15 +91,18 @@ export class L2TransactionReceipt implements TransactionReceipt {
...
+  }
 
-    return logs.map(
-      log => iface.parseLog(log).args as L2ToL1TransactionEvent['args']
-    )
+  private isClassic(
+    e: L2ToL1TransactionEvent
+  ): e is ClassicL2ToL1TransactionEvent['args'] {
+    if ((e as ClassicL2ToL1TransactionEvent['args']).indexInBatch) return true
+    else return false
```
fredlacs, line 105:
```suggestion
    return 'indexInBatch' in e
```

fredlacs (changes requested):

src/lib/message/L2ToL1Message.ts
```diff
@@ -65,104 +66,62 @@ export enum L2ToL1MessageStatus {
...
-        t.filters.L2ToL1Transaction(null, destination, uniqueId, batchNumber),
-      filter
+    const classicLogs = await classic.L2ToL1Message.getL2ToL1MessageLogs(
+      l2Provider,
+      filter,
+      positionOrBatchNumber,
+      destination,
+      hashOrUniqueId,
+      indexInBatch
     )
```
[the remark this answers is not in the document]
yahgwai: We dont have that block yet right? We can add that optimisation when it's available from Nitro node.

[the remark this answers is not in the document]
fredlacs: We could add it to the networks object, and check against that (something like 0 for devnet and 99999999..999 for testnet/Mainnet)

src/lib/message/L2ToL1Message.ts
```diff
@@ -65,104 +66,62 @@ export enum L2ToL1MessageStatus {
...
-
 /**
  * Base functionality for L2->L1 messages
  */
 export class L2ToL1Message {
+  protected isClassic(
+    e: L2ToL1TransactionEvent
+  ): e is ClassicL2ToL1TransactionEvent['args'] {
+    if ((e as ClassicL2ToL1TransactionEvent['args']).indexInBatch) return true
+    else return false
```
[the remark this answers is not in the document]
yahgwai: That looks clearer, but it doesnt offer any type safety. I've updated to:
```
return !!(e as ClassicL2ToL1TransactionEvent['args']).indexInBatch
```

src/lib/message/L2Transaction.ts
yahgwai: As above

src/lib/message/L2ToL1Message.ts
yahgwai: Keeping it as one parameter maintains the same API. I'll add some documentation about why they're named this way.

src/lib/message/L2ToL1Message.ts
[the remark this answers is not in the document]
fredlacs: fair, im game for typesafe but then lets not have the falsey edge case here heh
```javascript
return (e as ClassicL2ToL1TransactionEvent['args']).indexInBatch !== undefined
```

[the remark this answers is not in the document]
fredlacs: its a big number so there isnt the falsey of zero annoyance, nice

fredlacs (approved):
LGTM

PR #44, opened 2022-05-11
feat: Added ability to get deposits as transaction requests + some extra data

yahgwai:


DZGoldman:
^ wrong Daniel Of Offchain Labs :)

fredlacs (changes requested):

src/lib/message/L1Transaction.ts
```diff
@@ -238,12 +239,11 @@ export class L1TransactionReceipt implements TransactionReceipt {
       )
       .map(m => {
         const value = this.parseEthDepositData(m.inboxMessageEvent.data)
-
         return new EthDepositMessage(
           l2Provider,
           chainID,
           m.inboxMessageEvent.messageNum,
-          m.bridgeMessageEvent.sender,
+          new Address(m.bridgeMessageEvent.sender).applyAlias().value,
```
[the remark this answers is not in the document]
fredlacs: > I didnt change the from/to 
Agreed its fine as is, but would be good to have the alias logic inside `EthDepositMessage`

src/lib/dataEntities/transactionRequest.ts
```diff
@@ -0,0 +1,51 @@
...
+  Required<Pick<TType, TProps>> &
+    // and include all existing properies
+    TType
+
+/**
+ * A transaction request for a transaction that will trigger some sort of
+ * execution on the L2
+ */
+export interface L1ToL2TransactionRequest {
+  txRequest: NonOptional<TransactionRequest, 'to' | 'data' | 'value'>
```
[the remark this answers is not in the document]
yahgwai: They define the minimum for a user interaction - from these fields the others can be derived in different ways> What I wanted to do was have only those fields, but make sure they overlap with a transaction request. I think a better way and clearer to do this would be use `Pick` though - I'll update.

src/lib/dataEntities/transactionRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * A transaction request for a transaction that will trigger some sort of
+ * execution on the L2
+ */
+export interface L1ToL2TransactionRequest {
+  /**
+   * Core fields needed to forma transaction request: `to`, `data`, `from`
+   */
+  txRequestCore: Required<Pick<TransactionRequest, 'to' | 'data' | 'value'>>
```
fredlacs, line 12: could argue that the gas field also defines the behaviour, depends if you assume this request is pre or post estimation.
that said, instead of having gas here it seems like a fair assumption that its whatever amount of gas needed for the tx not to revert
```suggestion
   * Core fields needed to form a transaction request: `from`, `to`, `data`, `value`, 'chainId', 'nonce'
   */
  txRequestCore: Required<Pick<TransactionRequest, 'from' | 'to' | 'data' | 'value' | 'chainId' | 'nonce'>>
```

src/lib/dataEntities/transactionRequest.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A transaction request for a transaction that will trigger some sort of
+ * execution on the L2
+ */
+export interface L1ToL2TransactionRequest {
+  txRequest: NonOptional<TransactionRequest, 'to' | 'data' | 'value'>
+  /**
+   * The gas limit provided to this transactin when executed on L2 (units of gas)
+   */
+  l2GasLimit: BigNumber
```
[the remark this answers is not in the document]
fredlacs: I think we should think through the flow for validating these before merging this PR.

maybe something like the following to force consumers to think about validation
```ts
const requestValidator = lib.getDepositRequest(params)
const txRequest = requestValidator.getRequestIfValid()
```

src/lib/dataEntities/transactionRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+   * The L2 retryable ticket submission cost (wei).
+   * Caution: The Arbitrum submission price changes according to the l1 base fee,
+   * which in turn means that the l2SubmissionFee needs to be changed in order to ensure submission succeeds.
+   * If the l1 base fee increases by too much, then this transaction will fail upon execution on the L1.
+   */
+  l2SubmissionFee: BigNumber
+  /**
+   * The maximum total amount of eth that could be spent on L2 (wei)
+   */
+  l2GasCostsMaxTotal: BigNumber
```
fredlacs, line 35: is this supposed to be `l2MaxFeePerGas * l2GasLimit + l2SubmissionFee + l2Callvalue`?

src/lib/dataEntities/transactionRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  excessFeeRefundAddress: string
+  /**
+   * The address the call value will be refunded to in event of failure
+   */
+  callValueRefundAddress: string
+  /**
+   * The value of the retryable
+   */
+  value: BigNumber
```
fredlacs, line 65: do you mean the deposit amount? (ie how much msg.value was sent to the L1 inbox?)
this isn't necessarily the L2 call value

src/lib/dataEntities/transactionRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Ensure the T is not of TransactionRequest type by ensure it doesnt have a specific TransactionRequest property
+ */
+type IsNotTransactionRequest<T> = T extends { txRequest: any } ? never : T
```
fredlacs, line 71:
```suggestion
type IsNotTransactionRequest<T> = T extends { txRequestCore: any } ? never : T
```

src/lib/message/L1ToL2MessageCreator.ts
```diff
@@ -13,6 +13,43 @@ import { BigNumber } from 'ethers'
 import { SignerProviderUtils } from '../dataEntities/signerOrProvider'
 import { MissingProviderArbSdkError } from '../dataEntities/errors'
 import { getBaseFee } from '../utils/lib'
+import {
+  isL1ToL2TransactionRequest,
+  L1ToL2TransactionRequest,
+} from '../dataEntities/transactionRequest'
+
+interface L1ToL2MessageParams {
```
fredlacs, line 21: this seems like a mix of the params from `IRetryableData` and `L1ToL2TransactionRequest` from `src/lib/dataEntities/transactionRequest.ts`.
I don't mind the repeated code, but maybe there's a neat way to conciliate these 3 interfaces to share more stuff

src/lib/message/L1ToL2MessageCreator.ts
```diff
@@ -25,60 +62,85 @@ export class L1ToL2MessageCreator {
...
+        data: functionData,
+        value: estimates.totalL2GasCosts.add(params.l2CallValue),
+      },
+    }
+  }
+
+  /**
+   * Creates a retryable ticket by directly calling the Inbox contract on L1
+   */
+  public async createRetryableTicket(
```
fredlacs, line 129: I like this approach of having a separate function to get the request, another option we could take is have an override optional parameter which makes this same function return the request instead of submitting the tx
I guess the tradeoff is less discoverability when browsing the api, but makes it neater to toggle any tx into a request with minimal changes on the consumer side
(I don't know if it would necessarily be better, but bringing it up for discussion)

src/lib/message/L1Transaction.ts
```diff
@@ -365,7 +365,7 @@ export class L1EthDepositTransactionReceipt extends L1TransactionReceipt {
     const res = await message.wait(confirmations, timeout)
 
     return {
-      complete: !!res,
+      complete: Boolean(res),
```
fredlacs, line 368:
```suggestion
      complete: isDefined(res),
```

fredlacs (changes requested):

src/lib/dataEntities/transactionRequest.ts
yahgwai: Yeh, i had that originally. But the problem with extending the tx request is that ethers js gets angry with unexpected properties. So if you extend the tx request you can no longer do: `signer.sendTransaction(txRequest)`

fredlacs: what if we explicitly make L1ToL2TransactionRequest an intersection type with [ether's](https://github.com/ethers-io/ethers.js/blob/a71f51825571d1ea0fa997c1352d5b4d85643416/packages/abstract-provider/src.ts/index.ts#L18)?

src/lib/dataEntities/transactionRequest.ts
yahgwai: Ah no, it's just gas costs. I'll add that to the comment. I dont want to include the formula for calculating the gas costs, as that's an implementation detail, but it's just supposed to be the total amount of eth spent on gas on l2. In practice that's 
```
l2MaxFeePerGas * l2GasLimit + l2SubmissionFee
```
for retryables, and `0` for eth deposits.

src/lib/message/L1ToL2MessageCreator.ts
yahgwai: updated to be a function of retryable data, now a breaking change

src/lib/message/L1ToL2MessageCreator.ts
yahgwai: yeh, I think I like the discoverability. It will also make for more readable code for the client.

src/lib/dataEntities/transactionRequest.ts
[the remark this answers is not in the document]
yahgwai: Added isValid functionality

PR #55, opened 2022-05-25
feat: Converted provider functions to arbitrum provider

yahgwai:


src/lib/utils/arbProvider.ts
```diff
@@ -54,46 +55,38 @@ class ArbFormatter extends Formatter {
...
 
-  return includeTransactions
-    ? arbFormatter.blockWithTransactions(l2Block)
-    : (arbFormatter.block(l2Block) as unknown as ArbBlockWithTransactions)
+  public override async getTransactionReceipt(
+    transactionHash: string | Promise<string>
+  ): Promise<ArbTransactionReceipt> {
+    return (await super.getTransactionReceipt(
+      transactionHash
+    )) as ArbTransactionReceipt
```
DZGoldman, line 76: we could sanity check for the inclusion of the arb-specific fields before the type assertions here (and the other 2 functions below) (and maybe throw an error if not included? this would just be to prevent confusion if an ArbProvider isn't really an Arbitrum node, maybe unnecessary)

DZGoldman (approved):

PR #56, opened 2022-05-26
feat: Estimates for L1 costs of withdrawals, and tx requests for them

yahgwai:


src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -207,29 +218,75 @@ export class Erc20Bridger extends AssetBridger<
...
+    }
+  }
+
+  private isApproveParams(
+    params:
+      | TokenApproveParams
+      | {
+          txRequest: Required<Pick<TransactionRequest, 'to' | 'data' | 'value'>>
+          l1Signer: Signer
+        }
```
fredlacs, line 259: We could do something like 
```ts
      Parameters<typeof approveToken>
```

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -207,29 +218,75 @@ export class Erc20Bridger extends AssetBridger<
...
 
   /**
-   * Approve tokens for deposit to the bridge. The tokens will be approved for the relevant gateway.
+   * Get a tx request to approve tokens for deposit to the bridge.
+   * The tokens will be approved for the relevant gateway.
    * @param params
    * @returns
    */
-  public async approveToken(
+  public async getApproveTokenRequest(
```
fredlacs, line 226: getting the request currently requires a `l1Signer: Signer`
it would be cool to be able to instead provide a provider and from address.
The alternative would be to expect the consumer to use a [void signer](https://docs.ethers.io/v5/api/signer/#VoidSigner).

I think currently in deposits flow we instead surface the option to set a `from` address

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -640,27 +697,75 @@ export class Erc20Bridger extends AssetBridger<
...
+      throw new MissingProviderArbSdkError('l2Signer')
+    }
+    await this.checkL2Network(params.l2Signer)
+
+    const to = params.destinationAddress || (await params.l2Signer.getAddress())
+
+    const routerInterface = L2GatewayRouter__factory.createInterface()
+    const functionData =
+      // we need to do this since typechain doesnt seem to correctly create
+      // encodeFunctionData for functions with overrides
```
fredlacs, line 718: latest typechain version also has this bug? if so, would be cool if we opened an issue in their repo

src/lib/assetBridger/ethBridger.ts
```diff
@@ -133,23 +143,56 @@ export class EthBridger extends AssetBridger<
     return L1TransactionReceipt.monkeyPatchEthDepositWait(tx)
   }
 
+  public async getWithdrawalRequest(
+    params: EthWithdrawParams
```
fredlacs, line 147: same as before about signer

fredlacs (changes requested):

src/lib/assetBridger/erc20Bridger.ts
fredlacs: interesting it seems to be present in the export
https://unpkg.com/browse/@arbitrum/sdk@2.0.13/dist/lib/abi/L2GatewayRouter.d.ts
```ts
encodeFunctionData(functionFragment: "outboundTransfer", values: [string, string, BigNumberish, BytesLike]): string;
```

src/lib/assetBridger/erc20Bridger.ts
yahgwai: We dont actually need a from here since the actual approval will use msg.sender - so just replacing the l1signer with an l1provider and leaving at that

src/lib/assetBridger/erc20Bridger.ts
yahgwai: so the underlying contract has 2 functions on it both called "outboundTransfer" but with different args. So they need to be called with something like "outboundTransfer(address,address,uint256,bytes)" so that the underlying contract finds the correct one. The typechain interface only "outboundTransfer" - which isnt actually a func on the underlying ethers contract object.

Yeh, we need to create a ticket, or check if this is the case after upgrading.

src/lib/assetBridger/ethBridger.ts
yahgwai: updated here and on the erc20bridger

src/lib/assetBridger/erc20Bridger.ts
yahgwai: Yeh, looks like this is still an issue in the latest typechain

I'll create an issue

yahgwai: https://github.com/dethcrypto/TypeChain/issues/752

fredlacs (approved):
LGTM

PR #57, opened 2022-05-26
feat: Added retryable data parsing tools

yahgwai:


src/lib/dataEntities/retryableData.ts
```diff
@@ -0,0 +1,84 @@
...
+  gasLimit: BigNumber
+  maxFeePerGas: BigNumber
+  data: string
+}
+
+/**
+ * Tools for parsing retryable data from errors
+ */
+export class RetryableDataTools {
+  public static ErrorTriggeringParams = {
```
DZGoldman, line 21: feel (not strongly) like this should live in the tests, not the lib

DZGoldman (approved):

src/lib/dataEntities/retryableData.ts
yahgwai: I want these here so that people can use them to trigger the retryable data - added a better comment to explain this.

PR #58, opened 2022-05-26
chore: Speed up formatting by skipping the generated abi files

yahgwai:


DZGoldman (approved):
(this has always bugged me 👍 )

PR #59, opened 2022-05-26
feat: Added missing retryable functionality

yahgwai:


src/lib/message/L1ToL2Message.ts
```diff
@@ -548,14 +560,14 @@ export class L1ToL2MessageWriter extends L1ToL2MessageReader {
       )
     } else {
       throw new ArbSdkError(
-        `Cannot redeem. Message status: ${status} must be: ${L1ToL2MessageStatus.FUNDS_DEPOSITED_ON_L2}.`
+        `Cannot redeem as retryable does not exist. Message status: ${status} must be: ${L1ToL2MessageStatus.FUNDS_DEPOSITED_ON_L2}.`
```
DZGoldman, line 563: maybe `${status}` and `${L1ToL2MessageStatus.FUNDS_DEPOSITED_ON_L2}` would be useful as human-readable string here? (instead of integer)

DZGoldman (approved):

PR #60, opened 2022-05-31
Check LifetimeExtended event

gzeoneth:


src/lib/message/L1ToL2Message.ts
```diff
@@ -307,34 +308,36 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
         // We can skip by doing fromBlock.number + 1 on the first go
         // since creationBlock because it is covered by the `getAutoRedeem` shortcut
-        const redeemEventLogs = await this.l2Provider.getLogs({
-          fromBlock: fromBlock.number + 1,
-          toBlock: toBlockNumber,
-          topics: [redeemTopic, this.retryableCreationId],
-        })
-        const redeemEvents = redeemEventLogs.map(
-          r => iFace.parseLog(r).args as RedeemScheduledEvent['args']
+        queriedRange.push({ from: fromBlock.number + 1, to: toBlockNumber })
```
yahgwai, line 326: It's possible that the lifetime was extended in the first block, in which case I think we'd miss it.

src/lib/message/L1ToL2Message.ts
```diff
@@ -307,34 +308,37 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
     // for relevant redeem scheduled events
     if (creationReceipt) {
-      const iFace = ArbRetryableTx__factory.createInterface()
-      const redeemTopic = iFace.getEventTopic('RedeemScheduled')
-
       let increment = 1000
       let fromBlock = await this.l2Provider.getBlock(
         creationReceipt.blockNumber
       )
       const creationBlock = await this.l2Provider.getBlock(
```
yahgwai, line 315: I think you can get rid of the creationBlock var now

yahgwai (approved):

src/lib/message/L1ToL2Message.ts
```diff
@@ -345,12 +349,32 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+              break
+            }
+          }
+          if (toBlock.timestamp > timeout) break
+          // It is possible to have another keepalive in the last range as it might include block after previous timeout
+          while (queriedRange.length > 1) queriedRange.shift()
+        }
         const processedSeconds = toBlock.timestamp - fromBlock.timestamp
         if (processedSeconds != 0) {
           // find the increment that cover ~ 1 day
```
yahgwai, line 380: I think we should do:
```
increment = Math.ceil(increment * 86400 / processedSeconds)
```
instead of:
```
increment *= Math.ceil(86400 / processedSeconds)
```
since we're trying to approach the target of processedSeconds == 86400 the value of `Math.ceil(86400 / processedSeconds)` should always be either 1 or 2, which is quite a large jump. Rounding after multiplication will yield smaller increments.

gzeoneth: nice catch, pushed the fix

src/lib/message/L1ToL2Message.ts
gzeoneth: nice catch, pushed the fix

PR #63, opened 2022-06-09
fix: Nitro deposit format

yahgwai:
Update the l1 transaction receipt for eth deposits to follow the nitro format


fredlacs (approved):
LGTM
we should do a patch version bump here, right?

yahgwai:
> LGTM we should do a patch version bump here, right?

Yep, added version bump. We also need to publish the new nitro stable beta, and add that to the migration branch before publishing

PR #64, opened 2022-06-09
feat: handle undefined addresses with event fetcher

fredlacs:


src/lib/message/L1ToL2Message.ts
```diff
@@ -324,19 +320,23 @@ export class L1ToL2MessageReader extends L1ToL2Message {
 
         // We can skip by doing fromBlock.number + 1 on the first go
         // since creationBlock because it is covered by the `getAutoRedeem` shortcut
-        const redeemEventLogs = await this.l2Provider.getLogs({
```
yahgwai, line 327: I think this has already been updated on c-nitro-stable, we need to do a merge here.

src/lib/utils/eventFetcher.ts
```diff
@@ -57,17 +58,21 @@ export class EventFetcher {
     TContract extends Contract,
     TEventFilter extends TypedEventFilter<TypedEvent>
   >(
-    addr: string,
```
yahgwai, line 60: Maybe this should go in the filter objects. It should probably be an optional param - that will break backwards compatiblity though

PR #65, opened 2022-06-09
fix: correct eth deposit aliasing

yahgwai:


src/lib/message/L1Transaction.ts
```diff
@@ -239,11 +240,21 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
 
+        // we need to apply an alias to the sender to get the to address, but only if the sender was an EoA
+        // https://github.com/OffchainLabs/nitro/blob/a4eb505ae4f47967fa0ec18e85a43aa4602540c5/contracts/src/bridge/Inbox.sol#L216
+        const aliasedSender = new Address(
+          m.bridgeMessageEvent.sender
+        ).applyAlias().value
+        const to =
+          this.from.toLowerCase() === aliasedSender.toLowerCase()
+            ? aliasedSender
+            : m.bridgeMessageEvent.sender
```
fredlacs, line 251: can we have this logic live in `EthDepositMessage`?

src/lib/message/L1Transaction.ts
```diff
@@ -239,11 +240,21 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
+
         return new EthDepositMessage(
           l2Provider,
           chainID,
           m.inboxMessageEvent.messageNum,
-          m.bridgeMessageEvent.sender,
+          to,
```
fredlacs, line 257: I think we should refactor this to expect the `from` address (ie whoever called the L1 inbox.
We can then internally calculate the alias if needed (ie if we detect its a L1 contract)

fredlacs, line 257: I think we should refactor this to expect the `from` address (ie whoever called the L1 inbox.
We can then internally calculate the alias if needed (ie if we detect its a L1 contract)

fredlacs (changes requested):

src/lib/message/L1Transaction.ts
fredlacs: discussed offline, we can do this in a separate static entrypoint

src/lib/message/L1ToL2Message.ts
```diff
@@ -688,6 +701,73 @@ export class EthDepositMessage {
...
+   * @param inboxMessageEvent The InboxMessageDeliveredEvent emitted from the depositing transaction
+   * @param bridgeMessageEvent The MessageDeliveredEvent emitted from Bridge.sol in the depositing transaction
+   * @returns
+   */
+  public static fromTxReceipt(
+    l2Provider: Provider,
+    chainId: number,
+    txReceipt: TransactionReceipt,
+    inboxMessageEvent: InboxMessageDeliveredEvent['args'],
+    bridgeMessageEvent: MessageDeliveredEvent['args']
```
fredlacs, line 735: can't we get these values from the `txReceipt`?
would be cool to simplify the interface

src/lib/message/L1ToL2Message.ts
```diff
@@ -688,6 +701,73 @@ export class EthDepositMessage {
...
+   * @param l2Provider
+   * @param chainId
+   * @param txReceipt Receipt for the transaction originating the events
+   * @param inboxMessageEvent The InboxMessageDeliveredEvent emitted from the depositing transaction
+   * @param bridgeMessageEvent The MessageDeliveredEvent emitted from Bridge.sol in the depositing transaction
+   * @returns
+   */
+  public static fromTxReceipt(
+    l2Provider: Provider,
+    chainId: number,
```
fredlacs, line 732: we can simplify the api a bit by making this function async then inferring the chainId from the provider
```ts
    const chainID = (await l2Provider.getNetwork()).chainId
```

src/lib/message/L1ToL2Message.ts
fredlacs: if we don't do that, the check below isn't sufficient
```ts
if (!inboxMessageEvent.messageNum.eq(bridgeMessageEvent.messageIndex)) {
```
we also need to check if inboxMessageEvent was actually emitted in this given tx receipt

src/lib/message/L1ToL2Message.ts
```diff
@@ -673,6 +677,75 @@ export class EthDepositMessage {
...
+
+        // we need to apply an alias to the sender to get the to address, but only if the sender was an EoA
+        // https://github.com/OffchainLabs/nitro/blob/a4eb505ae4f47967fa0ec18e85a43aa4602540c5/contracts/src/bridge/Inbox.sol#L216
+        const aliasedSender = new Address(
+          m.bridgeMessageEvent.sender
+        ).applyAlias().value
+        const to =
+          txReceipt.from.toLowerCase() === aliasedSender.toLowerCase()
+            ? aliasedSender
+            : m.bridgeMessageEvent.sender
```
fredlacs, line 729: we should always apply the alias here since its the data that comes from the event

fredlacs (changes requested):

fredlacs (approved):
LGTM

PR #70, opened 2022-06-10
chore: move dependencies into dev dependencies

fredlacs:


fredlacs:
breaking change on network type to remove rpc url

scripts/instantiate_bridge.ts
```diff
@@ -85,14 +85,20 @@ export const instantiateBridge = (
...
-  if (!l1Network.rpcURL) {
-    throw new Error('L1 rpc url not set (see .env.sample or networks.ts)')
-  }
-  if (!l2Network.rpcURL) {
-    throw new Error('L2 rpc url not set (see .env.sample or utils/networks.ts)')
-  }
-  const ethProvider = new JsonRpcProvider(l1Network.rpcURL)
-  const arbProvider = new JsonRpcProvider(l2Network.rpcURL)
+  const l1Rpc = (() => {
+    if (networkID === 42161) return process.env['MAINNET_RPC'] as string
```
yahgwai, line 89: network id is always l2 network - could rename it to l2networkId

src/lib/dataEntities/networks.ts
```diff
@@ -135,7 +131,6 @@ export const l1Networks: L1Networks = {
     explorerUrl: 'https://etherscan.io',
     partnerChainIDs: [42161],
     blockTime: 14,
-    rpcURL: process.env['MAINNET_RPC'] as string,
```
yahgwai, line 138: I've got a feeling these are used in the token-bridge-ui, we should take a look there

scripts/instantiate_bridge.ts
```diff
@@ -85,14 +85,20 @@ export const instantiateBridge = (
...
+    if (networkID === 421611) return process.env['RINKEBY_RPC'] as string
+    if (networkID === 1338) return 'http://127.0.0.1:8545/'
```
yahgwai, line 91: What will we do for unknown networks, like custom networks, local networks on different chain id or newly created devnets?

src/lib/dataEntities/networks.ts
fredlacs: yea but to instantiate a network you give a provider (in getNetwork)
so they already have the rpc lying around

scripts/instantiate_bridge.ts
fredlacs: they can have their own `instantiate_bridge` script or we can add support when needed.
this script isn't published as part of the package, its a dev utility

src/lib/dataEntities/networks.ts
fredlacs: I don't think its useful for us to have the rpc url here since its never used by the sdk package itself

fredlacs: we should be good
https://github.com/OffchainLabs/arb-token-bridge/search?q=rpcURL

fredlacs: cc @spsjvc

yahgwai (approved):

PR #75, opened 2022-06-10
Fix incorrect comparison

haryu703:
Modify to use the [specified method](https://docs.ethers.io/v5/api/utils/bignumber/#BigNumber--BigNumber--methods--comparison-and-equivalence
) for comparison between BigNumbers.

Using `>=` may return incorrect values.
```node
> a
BigNumber { _hex: '0x2386f26fc10000', _isBigNumber: true }
> b
BigNumber { _hex: '0x3d570965c1', _isBigNumber: true }
> a >= b
false
> a.gte(b)
true
```


gzeoneth:
Thanks @haryu703, I have the same fix in #76 and will merge that instead due to the CLA stuff.

PR #78, opened 2022-06-13
chore: add tslint strict-comparisons and no-implicit-coercion

gzeoneth:
Adding [tslint strict-comparisons rule](https://palantir.github.io/tslint/rules/strict-comparisons/) to detect issue similar to #76. Using [tslint within eslint](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin-tslint) because the required rule is not ported https://github.com/typescript-eslint/typescript-eslint/pull/314. Since this rule required parserServices to be generated, only files included in `tsconfig.json` (i.e. src/**/*.ts) can be linted.

Using comparison operators with BigNumber will have undesired behavior, e.g.
```
> node 
Welcome to Node.js v16.14.2.
Type ".help" for more information.
> const bn = (await import("ethers")).BigNumber
undefined
> bn.from(1) < bn.from(2)
true
> bn.from(9) < bn.from(10)
false
```

This PR also fix a minor audit-ci issue with 1702d6a.

h/t @haryu703 bringing this up in Discord and submited PR #75


fredlacs:
I think we should also add the [no-implicit-coercion](https://eslint.org/docs/rules/no-implicit-coercion) rule

yahgwai (approved):

PR #82, opened 2022-06-15
feat: Add the legacy getInputs function

yahgwai:


gzeoneth (commented):
missing `getInputs()` in `IL1ToL2MessageReader` making it only usable by 
```
await (message as L1ToL2MessageReader).getInputs()
```

gzeoneth:
consider also merge #85 before release

fredlacs:
could we update the classic package to 1.1.9?

yahgwai:
> could we update the classic package to 1.1.9?

Updated

src/lib/message/L1Transaction.ts
```diff
@@ -184,13 +198,7 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
     messageIndex?: number
   ): Promise<IL1ToL2MessageReader | IL1ToL2MessageWriter> {
-    return (await isNitroL2(l2SignerOrProvider))
-      ? L1ToL2Message.fromNitro(
-          (await this.nitroReceipt.getL1ToL2Messages(l2SignerOrProvider))[
-            messageIndex || 0
-          ]
-        )
-      : this.classicReceipt.getL1ToL2Message(l2SignerOrProvider, messageIndex)
+    return (await this.getL1ToL2Messages(l2SignerOrProvider))[messageIndex || 0]
```
gzeoneth, line 201: This function is differ from the spec
>   * If the messageIndex is supplied the message at that index will be returned.
>   * If no messageIndex is supplied a message will be returned if this transaction only created one message
>   * All other cases throw an error

It return `undefined` if the message is not found instead of throw

gzeoneth: It also don''t throw when no messageIndex is supplied and the transaction created more than 1 message

src/lib/message/L1ToL2Message.ts
```diff
@@ -234,6 +296,30 @@ export class L1ToL2MessageWriter
...
+      classicWriter.retryableCreationId
+    )
+  }
+
+  public static fromNitro(nitroWriter: nitro.L1ToL2MessageWriter) {
+    return new L1ToL2MessageWriter(
+      nitroWriter.l2Signer,
+      nitroWriter.chainId,
+      nitroWriter.sender,
+      undefined,
```
gzeoneth, line 316: L1ToL2MessageReader constructor require `messageNumber`

gzeoneth:
```suggestion
      nitroWriter.messageNumber,
```

gzeoneth:
Why don't we return the callData (but only callDataLength) here?

src/lib/message/L1ToL2Message.ts
yahgwai: good catch

yahgwai:
> Why don't we return the callData (but only callDataLength) here?

I think that's just what we used to do in arb-ts, i cant really remember. This function wont exist in nitro though.

src/lib/message/L1Transaction.ts
yahgwai: Updated

gzeoneth (approved):
LGTM

PR #86, opened 2022-06-16
feat: add support for new eth deposit msg and include status function

fredlacs:
adds support for changes in https://github.com/OffchainLabs/go-ethereum/pull/112
also adds a `status` function you can query on eth deposits

### DIFF
--- a/src/lib/message/L1Transaction.ts
+++ b/src/lib/message/L1Transaction.ts
@@ -220,7 +220,7 @@ export class L1TransactionReceipt implements TransactionReceipt {
   public async getEthDepositMessages(
     l2Provider: Provider
   ): Promise<EthDepositMessage[]> {
-    return await Promise.all(
+    return Promise.all(
       this.getMessageEvents()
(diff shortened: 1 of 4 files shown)


fredlacs:
ah we prob need to bump up the nitro version used in the integration tests

src/lib/message/L1Transaction.ts
```diff
@@ -217,10 +217,18 @@ export class L1TransactionReceipt implements TransactionReceipt {
     }
   }
 
+  public async getAllDeposits<T extends SignerOrProvider>(l2SignerOrProvider: T) {
```
yahgwai, line 220: I don't think this adds much value, a caller can do this themselves if they want both. 

EthDepositMessage and L1ToL2Message are quite different objects, so in most cases a caller will need to be filtering these objects anyway, checking the type, then doing different actions for each type.

fredlacs: removed

PR #87, opened 2022-06-16
fix: surface eth deposits in migration package

fredlacs:
When using sdk v1 users would get the eth deposit message when interacting with the migration sdk
With the nitro sdk v3, you need a separate function call to get eth deposits.

We should bump up the nitro package to include https://github.com/OffchainLabs/arbitrum-sdk/pull/86 so the `status` function is also available

We could attempt to make `EthDepositMessage` closer to a `IL1ToL2MessageReaderOrWriter` but most methods would throw errors since they don't make sense.

Instead making a separate entrypoint function `getAllDeposits`, need to publish new nitro beta version to consume this here though


fredlacs:
closed in favour of https://github.com/OffchainLabs/arbitrum-sdk/pull/88

PR #89, opened 2022-06-20
fix: validate l1 address before returning

yahgwai:
We make the l1provider optional so as not to break the api

### DIFF
--- a/src/lib/assetBridger/erc20Bridger.ts
+++ b/src/lib/assetBridger/erc20Bridger.ts
@@ -364,7 +364,7 @@ export class Erc20Bridger extends AssetBridger<
 
     // @deprecated: for the next breaking change make the l1Provider required and always do the check below
     if (l1Provider) {
-      await this.checkL1Network(l2Provider)
+      await this.checkL1Network(l1Provider)
       // check that this l1 address is indeed registered to this l2 token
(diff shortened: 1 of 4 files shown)


src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -338,18 +348,38 @@ export class Erc20Bridger extends AssetBridger<
...
+    l2Provider: Provider,
+    l1Provider?: Provider
   ): Promise<string> {
     await this.checkL2Network(l2Provider)
 
     const arbERC20 = L2GatewayToken__factory.connect(erc20L2Address, l2Provider)
+    const l1Address = await arbERC20.functions.l1Address().then(([res]) => res)
+
+    // @deprecated: for the next breaking change make the l1Provider required and always do the check below
+    if (l1Provider) {
```
fredlacs, line 366: we can instead just check against the L2 router using the L2 provider

fredlacs (changes requested):

src/lib/assetBridger/erc20Bridger.ts
yahgwai: good point. updated

fredlacs (approved):
LGTM

PR #94, opened 2022-06-21
fix: improve getL1ERC20Address error message

gzeoneth:
For token without .l1Address() (e.g. DAI), `getL1ERC20Address` will throw with CALL_EXCEPTION

https://github.com/OffchainLabs/arbitrum-sdk/blob/afd58837a06f4e061b8f56f26db14e50439c7d7b/src/lib/assetBridger/erc20Bridger.ts#L355-362
```solidity
  public async getL1ERC20Address(
    erc20L2Address: string,
    l2Provider: Provider
  ): Promise<string> {
    await this.checkL2Network(l2Provider)

    const arbERC20 = L2GatewayToken__factory.connect(erc20L2Address, l2Provider)
    const l1Address = await arbERC20.functions.l1Address().then(([res]) => res)
```

There doesn't seems to be any on-chain registry that can resolve a L2 -> L1 address mapping. Another question is do we want to special case DAI by hardcoding the address here (but still do the validation to make it less risky).

Also added a validation fallback in [8b5467d](https://github.com/OffchainLabs/arbitrum-sdk/pull/94/commits/8b5467de95176db05755b99af1a253c827760ed9).

### DIFF
--- a/src/lib/assetBridger/erc20Bridger.ts
+++ b/src/lib/assetBridger/erc20Bridger.ts
@@ -364,7 +364,24 @@ export class Erc20Bridger extends AssetBridger<
     try {
       l1Address = (await arbERC20.functions.l1Address())[0]
     } catch (error) {
-      throw new ArbSdkError(`Unable to get l1 address of ${erc20L2Address}`)
+      if (erc20L2Address.toLowerCase() === '0xda10009cbd5d07dd0cecc66161fc93d7c9000da1'){
+        // hardcoding DAI here since it does not implement IArbToken
+        const chainId = (await l2Provider.getNetwork()).chainId
+        if (chainId === 42161){
+          // arbitrum mainnet
+          // the l1Address will still be validated with router
+          l1Address = '0x6b175474e89094c44da98b954eedeac495271d0f'
+        } else if (chainId === 421611){
+          // arbitrum rinkeby
+          // note that DAI's custom gateway was not registered to our router yet
+          // this l1Address will fail the validation
+          l1Address = '0xd9e66a2f546880ea4d800f189d6f12cc15bff281'
+        } else {
+          throw new ArbSdkError(`Unable to get l1 address of ${erc20L2Address}`)
+        }
+      } else {
+        throw new ArbSdkError(`Unable to get l1 address of ${erc20L2Address}`)
+      }
     }
 
     // check that this l1 address is indeed registered to this l2 token
--- a/src/lib/assetBridger/erc20Bridger.ts
+++ b/src/lib/assetBridger/erc20Bridger.ts
@@ -363,14 +363,17 @@ export class Erc20Bridger extends AssetBridger<
     try {
       l1Address = (await arbERC20.functions.l1Address())[0]
     } catch (error) {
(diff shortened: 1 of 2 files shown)


src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -369,9 +375,19 @@ export class Erc20Bridger extends AssetBridger<
 
     const l2Address = await l2GatewayRouter.calculateL2TokenAddress(l1Address)
     if (l2Address.toLowerCase() !== erc20L2Address.toLowerCase()) {
-      throw new ArbSdkError(
-        `Unexpected l1 address. L1 address from token is not registered to the provided l2 address. ${l1Address} ${l2Address} ${erc20L2Address}`
+      // also check with the standard gateway
```
yahgwai, line 378: Why also check with the standard gateway? Is this because we want to not throw if the l2address is the standard token, but a custom token has already been registered over the top?

yahgwai:
> There doesn't seems to be any on-chain registry that can resolve a L2 -> L1 address mapping. Another question is do we want to special case DAI by hardcoding the address here (but still do the validation to make it less risky).

The problem with hardcoding is that it wont work for new networks, and custom networks. Do you have a link to the L2 Dai contract, why doesnt it follow our L2 token standard?

src/lib/assetBridger/erc20Bridger.ts
gzeoneth: That is what I am thinking, would make bridging back those token to L1 easier. Or do you think we should simply throw?

gzeoneth:
> > There doesn't seems to be any on-chain registry that can resolve a L2 -> L1 address mapping. Another question is do we want to special case DAI by hardcoding the address here (but still do the validation to make it less risky).
> 
> The problem with hardcoding is that it wont work for new networks, and custom networks. Do you have a link to the L2 Dai contract, why doesnt it follow our L2 token standard?

It won't work for new networks, but since we validate the address it won't bring unsafe behavior either.
They simply don't implement our IArbToken interface.
https://arbiscan.io/address/0xda10009cbd5d07dd0cecc66161fc93d7c9000da1#code

src/lib/assetBridger/erc20Bridger.ts
yahgwai: Yeh, I think what you have makes sense. We're trying to protect against fraud here I guess, and that's not possible with the standard token. 

I think just add more comments, it took me a bit of thinking to figure out why we wanted to do this

yahgwai:
> It won't work for new networks, but since we validate the address it won't bring unsafe behavior either. They simply don't implement our IArbToken interface. https://arbiscan.io/address/0xda10009cbd5d07dd0cecc66161fc93d7c9000da1#code

Yeh ok, lets add the hardcoded addresses. Not perfect but it helps a bit.

src/lib/assetBridger/erc20Bridger.ts
gzeoneth: On second thought I think we might just ignore this and throw. Even if we let it pass validation here it still need to use the StandardGateway manually, instead of the router.

yahgwai (approved):

gzeoneth:
bumped to 1.1.10, @yahgwai can you merge and release?

PR #98, opened 2022-06-24
Fix new Nitro changes

gzeoneth:
https://github.com/OffchainLabs/nitro/pull/623 Outbox replay state management
https://github.com/OffchainLabs/nitro/pull/692 Refactor accumulator flow
https://github.com/OffchainLabs/nitro/pull/697 Move address aliasing from L2 to L1
https://github.com/OffchainLabs/nitro/pull/720 Expose isSpent in Outbox

Also fix a race condition in `L1ToL2Message.receiptsToStatus`


src/lib/message/L1ToL2Message.ts
```diff
@@ -432,7 +429,20 @@ export class L1ToL2MessageReader extends L1ToL2Message {
     }
 
     // not redeemed, has it now expired
-    if (await this.isExpired()) {
+    try {
+      if (await this.isExpired()) {
```
yahgwai, line 433: I think we should make `isExpired` atomic with `getSuccessfulRedeem` so that we always have consistent results for `getSuccessfulRedeem`.

We could adjust `isExpired` so that it has a try/catch internally, and rename it to `retryableExists`.

Then adjust `getSuccessfulRedeem` so that we call `retryableExists` somewhere at the start of it. If `retryableExists` is true then we can just return null straight away - since it hasnt been redeemed or expired. If `retryableExists` is false then we either successfully redeemed or we expired. We then look for a successful redeem, if we find it we return the tx receipt, if we dont we return some value that indicates an expiration.

Then we won't check `isExpired/retryableExists` in receiptToStatus, instead we just pass in the receipt, or the value that indicates expiration.

yahgwai (commented):
We need to PR this against `c-nitro-stable`. We could merge it into `c-nitro-next` and then cherry-pick it over to `c-nitro-stable`

src/lib/message/L1ToL2Message.ts
```diff
@@ -127,16 +127,13 @@ export abstract class L1ToL2Message {
       return ethers.utils.stripZeros(value.toHexString())
     }
 
-    const addressAlias = new Address(fromAddress)
-
-    const from = addressAlias.applyAlias()
```
fredlacs, line 132: sure we don't need to apply an alias here anymore?
If the user is supplying the `from` address as their L1 address it still needs to be aliased.

The difference is that now the alias is applied int he L1 inbox instead of by arb-os.
We should clarify what's expected in the `fromAddress` field. If its the value from the Inbox delivered event, it doesn't need aliasing. If its the plain user address, it needs alaising

gzeoneth: ah, `calculateSubmitRetryableId` is public static
I think we should just add some comment regarding the input expected.

package.json
```diff
@@ -51,17 +51,13 @@
...
   },
   "devDependencies": {
-    "ts-node": "^10.2.1",
-    "dotenv": "^10.0.0",
-    "@typechain/ethers-v5": "9.0.0",
-    "@types/yargs": "^17.0.9",
-    "typechain": "7.0.0",
-    "@arbitrum/nitro-contracts": "1.0.0-beta.5",
-    "arb-bridge-peripherals": "1.0.10",
+    "@arbitrum/nitro-contracts": "https://gitpkg.now.sh/OffchainLabs/nitro/contracts?expose-outbox-isspent",
```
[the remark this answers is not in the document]
gzeoneth: https://github.com/OffchainLabs/nitro/pull/723

PR #102, opened 2022-06-28
fix: Classic eth deposits bug - getInputs

yahgwai:


src/lib/message/L1Transaction.ts
```diff
@@ -104,13 +107,65 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
+      getAddress(hexZeroPad(bn.toHexString(), 20))
+    const destAddress = addressFromBigNumber(parsed[0])
+    const l2CallValue = parsed[1] as BigNumber
+    const l1Value = parsed[2] as BigNumber
+    const maxSubmissionFee = parsed[3] as BigNumber
+    const excessFeeRefundAddress = addressFromBigNumber(parsed[4])
+    const callValueRefundAddress = addressFromBigNumber(parsed[5])
+    const gasLimit = parsed[6] as BigNumber
+    const maxFeePerGas = parsed[7] as BigNumber
+    const data = '0x' + eventData.substring(64 * 9 + 2)
```
fredlacs, line 157: don't we want to use the parsed data length here?
something like
```ts
const dataLength = parsed[8] as BigNumber
const data = '0x' + eventData.substring(
    eventData.length - dataLength.toNumber()
)
```
I think we'd also need to adjust `eventData.length - dataLength.toNumber()` to take into account padding

fredlacs (approved):
LGTM

PR #103, opened 2022-06-28
Cherry pick message data parser from migration branch

yahgwai:


integration_test/messageDataParser.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      res.maxFeePerGas.eq(BigNumber.from('0x172c5865')),
+      'incorrect max fee per gas'
+    ).to.be.true
+    expect(
+      res.maxSubmissionFee.eq(BigNumber.from('0x53280cf149')),
+      'incorrect max submission fee'
+    ).to.be.true
+  })
+
+  it('does parse eth deposit message', async () => {
```
fredlacs, line 47: We should update this to parse nitro's Eth deposit message

fredlacs (approved):
LGTM

integration_test/messageDataParser.test.ts
yahgwai: Yeh, I left the old one there since that's still a valid kind of l1 to l2 message. But we can add some tests for the actual eth deposit message soon too.

I'll update the comment here

PR #106, opened 2022-06-30
feat: add flag to get unconfirmed outbox proof

gzeoneth:


src/lib/message/L2ToL1Message.ts
```diff
@@ -298,7 +303,7 @@ export class L2ToL1MessageReader extends L2ToL1Message {
     return this.l1BatchNumber
   }
 
-  protected async getSendProps(l2Provider: Provider) {
+  protected async getSendProps(l2Provider: Provider, requireConfirmed = true) {
```
yahgwai, line 306: I think a better way to do this would be to always look for the `latestNodeCreated` in getSendProps, but also store an isConfirmed flag for the resulting sendRootHash. You can populate isConfirmed by checking that `latestConfirmed` > `nodeNum`

I'm also ok with the way you have it, I think this flow is already a bit complicated which is why I'm suggesting the change above. If you decided to stick with the `requireConfirmed` flow I would suggest we make it non optional - typescript optional params can be a bit annoying around ordering, and I think it's generally nicer to be explicit by making it non optional.

gzeoneth: made it always look for the `latestNodeCreated` and added the `sendRootConfirmed` field

src/lib/message/L2ToL1Message.ts
```diff
@@ -299,30 +300,49 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
-        this.sendRootHash = l2Block.sendRoot
+      const sendRootSizeConfirmed = BigNumber.from(l2BlockConfirmed.sendCount)
+      if (sendRootSizeConfirmed.gt(this.event.position)) {
+        this.sendRootSize = sendRootSizeConfirmed
+        this.sendRootHash = l2BlockConfirmed.sendRoot
+        this.sendRootConfirmed = true
+      } else {
+        // Check latest (possibly unconfirmed) node
+        const latestNodeNum = await rollup.callStatic.latestNodeCreated()
+        if (latestNodeNum.gt(latestNodeNumConfirmed)) {
```
yahgwai, line 326: Why do we check this here? Isnt `latestNodeNum` always greater `latestNodeNumConfirmed` since `latestNodeCreated` is always `latestConfirmed`?

gzeoneth: practically yes but they can technically equal

gzeoneth: especially in AnyTrust?

yahgwai: Ok, i think we should include a comment about that then

PR #109, opened 2022-07-11
Updated tests to work against public testnets

yahgwai:


package.json
```diff
@@ -45,7 +47,7 @@
     "bridgeStandardToken": "ts-node scripts/deployStandard.ts"
   },
   "dependencies": {
-    "@arbitrum/sdk-classic": "npm:@arbitrum/sdk@1.1.10",
+    "@arbitrum/sdk-classic": "npm:@arbitrum/sdk@1.1.11",
```
fredlacs, line 50:
```suggestion
    "@arbitrum/sdk-classic": "npm:@arbitrum/sdk@1.1.12",
```

integration_test/standarderc20.test.ts
```diff
@@ -165,7 +169,7 @@ describe('standard ERC20', () => {
     await redeemAndTest(testState.l2Signer.provider!, waitRes.message, 1)
   })
 
-  it('deposit with low funds, fails first redeem, succeeds seconds', async () => {
+  it.skip('deposit with low funds, fails first redeem, succeeds seconds', async () => {
```
fredlacs, line 172: whats the reason to skip this?

yahgwai: I havent found a way to reliably get this test to pass now that the l1 gas price moves around, and slowly. This because in order for this test to pass we need to set a gas limit which is enough to pass the `redeem` function, but not enough for the actual scheduled redeem to work. I guess we could calculate this dynamically by looking at the l1 gas price estimate, but havent done that yet. I should add a ticket for this

scripts/testSetup.ts
```diff
@@ -184,6 +190,33 @@ export const getSigner = (provider: JsonRpcProvider, key?: string) => {
...
+  } else throw new ArbSdkError('Could not infer shadow networks.')
+
+  const l1Network = await nitroGetL1Network(l2Network.partnerChainID)
+  const copiedNetworks: { l1Network: L1Network; l2Network: L2Network } = {
+    l1Network: { ...l1Network },
+    l2Network: { ...l2Network },
+  }
+
+  copiedNetworks.l2Network.chainID = (await l2Provider.getNetwork()).chainId
+  copiedNetworks.l2Network.isCustom = true
```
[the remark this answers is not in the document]
yahgwai: updated

fredlacs (approved):
LGTM

PR #111, opened 2022-07-13
fix getFirstExecutableBlock

DZGoldman:


fredlacs (changes requested):
good catch, we should do the same change further above in
```ts
    if (proof === null)
      return BigNumber.from(network.confirmPeriodBlocks)
        .add(ASSERTION_CREATED_PADDING)
        .add(ASSERTION_CONFIRMED_PADDING)
```

DZGoldman:
@fredlacs done

yahgwai (approved):

src/lib/message/L2ToL1Message.ts
```diff
@@ -344,10 +344,13 @@ export class L2ToL1MessageReader extends L2ToL1Message {
...
+    if (proof === null) {
+      const latestBlock = await this.l1Provider.getBlockNumber()
       return BigNumber.from(network.confirmPeriodBlocks)
         .add(ASSERTION_CREATED_PADDING)
         .add(ASSERTION_CONFIRMED_PADDING)
+        .add(latestBlock)
+    }
     // we can't check if the L2 to L1 tx isSpent on the outbox, so we instead try executing it
     if (await this.hasExecuted(proof)) return BigNumber.from(0)
     const latestBlock = await this.l1Provider.getBlockNumber()
```
fredlacs, line 356: I think we're adding undefined on the previous block

fredlacs (changes requested):

src/lib/message/L2ToL1Message.ts
fredlacs: oops nevermind me, didn't see it inside the if

DZGoldman: eh?

fredlacs (approved):
LGTM for real now (but a small comment)

src/lib/message/L2ToL1Message.ts
```diff
@@ -344,10 +344,13 @@ export class L2ToL1MessageReader extends L2ToL1Message {
 
     const proof = await this.tryGetProof(l2Provider)
     // here we assume the L2 to L1 tx is actually valid, so the user needs to wait the max time.
-    if (proof === null)
+    if (proof === null) {
+      const latestBlock = await this.l1Provider.getBlockNumber()
```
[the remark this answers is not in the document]
DZGoldman: functionality is slightly more optimal this way (when already executed, saves a blocknum query!)

PR #113, opened 2022-07-14
Before and after tests for migration

yahgwai:


migration_test/migration.test.ts
```diff
@@ -0,0 +1,637 @@
...
+  Erc20Bridger,
+  IL2ToL1MessageWriter,
+} from '../src'
+dotenv.config()
+
+const waitForNitro = async (l2Provider: Provider, minimum?: number) => {
+  const now = Date.now()
+
+  while (
+    !((await isNitroL2(l2Provider)) && (!minimum || Date.now() - now > minimum))
```
fredlacs, line 45: we should probably try/catch this
if the sequencer is down during the migration, these inside `isNitroL2`
```ts
    const l2Network = await nitro.getL2Network(l2SignerOrProvider)
    const blockNumber = await arbSys.arbBlockNumber()
```

migration_test/migration.test.ts
```diff
@@ -0,0 +1,637 @@
...
+  }
+}
+
+abstract class MigrationTest {
+  public name = ''
+  constructor(public readonly testState: TestState) {}
+
+  public abstract initialise(): Promise<void>
+  public abstract finalise(): Promise<void>
+}
```
fredlacs, line 58: What is the value of using a class here?
I think the code will be cleaner and less dense if we instead declare an object
```ts
interface MigrationTest {
    name: string;
    testState: TestState;
    initialise: () => Promise<void>;
    finalise: () => Promise<void>;
}
```
then MultiTest can receive multiple `MigrationTest` objects

migration_test/migration.test.ts
```diff
@@ -0,0 +1,637 @@
...
+        erc20Bridger,
+        ethBridger,
+      },
+      ethTransfer: {},
+      tokenDeploy: {},
+      weth: {},
+      wethXChainDeposit: {},
+      wethXChainWithdrawal: {},
+      ethXChainWithdrawal: {},
+      ethXChainDeposit: {},
```
fredlacs, line 615: instead of having many empty things in the test state, we could instead just have `core` be part of `TestState`
then, we have the `initialise` function return a `context` object which is used as a param into the `finalize` function

fredlacs (changes requested):

fredlacs:
discussed offline: would be cool to have the weth deposit fail with gas limit of 0 then redeem after the migration is done

migration_test/migration.test.ts
yahgwai: updated

yahgwai:
Updated the weth test to manual redeem

fredlacs (approved):
LGTM

PR #121, opened 2022-07-25
Yarn audit allowlist updates

yahgwai:


fredlacs (approved):
[1081496](https://www.npmjs.com/advisories/1081496), [1081697](https://www.npmjs.com/advisories/1081697), [1081677](https://www.npmjs.com/advisories/1081677)
These are from hardhat as a dev dependency and shouldn't be a issue



[1081677](https://www.npmjs.com/advisories/1081680), [1081682](https://www.npmjs.com/advisories/1081682), [1081679](https://www.npmjs.com/advisories/1081679), [1081681](https://www.npmjs.com/advisories/1081681)
These are from OZ smart contracts (which are only used here for gen'ing ABIs so also shouldnt be an issue)

In the future we can remove the hardhat dependency from the packages, but we'll need to keep ignoring the OZ ones

PR #125, opened 2022-07-27
update nitro sdk version

DZGoldman:


yahgwai:
We've updated on another branch now

PR #134, opened 2022-08-01
Binary search in getFirstExecutableBlock

gzeoneth:
Calling L2ToL1Message.status() is still slow because `getSendProps` took 15s on my network for an unconfirmed message.
https://github.com/OffchainLabs/arbitrum-sdk/blob/105bf73cb788231b6e63c510713f460b36699fcd/src/lib/message/L2ToL1Message.ts#L216

Also can we narrow down the range of node to search before doing binary search? I guess we can binary search the L1 get log instead.


fredlacs (commented):
LGTM but lets give @yahgwai a chance to review before we merge

yahgwai (approved):
LGTM

PR #143, opened 2022-08-09
Added an override function for the partner chain id

yahgwai:


fredlacs (approved):
LGTM

gzeoneth (changes requested):
I think we can drop `lastUpdatedL1` and cache `isNitro` and `lastUpdatedL2` per L2 chain ID (may be in the network entity object), then in `isNitroL1()` we first get the partner L2 and use its `lastUpdatedL2` so we can avoid a lot of rpc call to `isNitro` functions when you switch between network frequently.

gzeoneth (approved):

fredlacs:
this version got already published, but good concerns raised.
lets address in a follow up PR?

PR #149, opened 2022-08-18
Remove hardcoded retryable data functions

yahgwai:


src/lib/message/L1ToL2MessageGasEstimator.ts
```diff
@@ -253,4 +256,84 @@ export class L1ToL2MessageGasEstimator {
...
+  public async populateL1ToL2FunctionParams(
+    /**
+     * Function that will internally make an L1->L2 transaction
+     * Will initially be called with dummy values to trigger a special revert containing
+     * the real params. Then called again with the real params to form the final data to be submitted
+     */
+    dataFunc: (params: {
+      gasLimit: BigNumber
+      maxFeePerGas: BigNumber
+      maxSubmissionFee: BigNumber
```
fredlacs, line 277: any reason not to use `L1ToL2MessageGasParams` as the type here?

src/lib/message/L1ToL2MessageGasEstimator.ts
```diff
@@ -253,4 +256,84 @@ export class L1ToL2MessageGasEstimator {
...
+    })
+
+    // get retryable data from the null call
+    const res = await l1Provider.call({
+      to: to,
+      data: nullData,
+      value: value,
+      from: from,
+    })
+    const retryable = RetryableDataTools.tryParseError(res)!
```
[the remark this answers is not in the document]
fredlacs: seems like ethersjs doesn't throw an error on solidity custom errors

src/lib/message/L1ToL2MessageGasEstimator.ts
yahgwai: No, reused now

src/lib/message/L1ToL2MessageGasEstimator.ts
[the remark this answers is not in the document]
yahgwai: I've wrapped this in a try/catch anyway, so if in future it does throw properly we can still catch here

src/lib/message/L1ToL2MessageGasEstimator.ts
```diff
@@ -253,4 +261,93 @@ export class L1ToL2MessageGasEstimator {
...
+        data: nullData,
+        value: value,
+        from: from,
+      })
+      retryable = RetryableDataTools.tryParseError(res)
+      if (!isDefined(retryable)) {
+        throw new ArbSdkError(`No retryable data found in error: ${res}`)
+      }
+    } catch (err) {
+      retryable = RetryableDataTools.tryParseError(err as Error)
```
fredlacs, line 311: would be good to add a comment explaining https://github.com/OffchainLabs/arbitrum-sdk/pull/149#discussion_r952677274

fredlacs (approved):
would be good to add a comment in for future reference, but lgtm

PR #150, opened 2022-08-22
Populate the l1rpc if it isnt populated already, used in the tutorials

yahgwai:
This is only a problem with the migration package, where isNitroL2 needs to get l1 rpc url, so we put onto the network object with an env var

### DIFF
--- a/src/lib/dataEntities/networks.ts
+++ b/src/lib/dataEntities/networks.ts
@@ -29,6 +29,7 @@ import {
   isNitroL1,
   isNitroL2,
 } from '../utils/migration_types'
+import { isDefined } from '../utils/lib'
 
 dotenv.config()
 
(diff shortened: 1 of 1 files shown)


GreatSoshiant (approved):
LGTM!

src/lib/dataEntities/networks.ts
```diff
@@ -91,11 +92,20 @@ export const getL1Network = async (
...
-  } else {
-    return await classic.getL1Network(signerOrProvider)
-  }
+  const network = await (async () => {
+    if (await isNitroL1(l2ChainId, signerOrProvider)) {
+      return await nitro.getL1Network(signerOrProvider)
+    } else {
+      return await classic.getL1Network(signerOrProvider)
+    }
+  })()
```
fredlacs, line 101:
```suggestion
  const network =
      ((await isNitroL1(l2ChainId, signerOrProvider)) ? nitro : classic)
          .getL1Network(signerOrProvider)
```

fredlacs (approved):
LGTM

PR #151, opened 2022-08-22
Network generation in L2ToL1Message

### LINKED ISSUE #132: Nitro network is not generated when using L2ToL1Message directly
In most flow, we call `isNitroL1` or `isNitroL2` in the migration sdk to detect if the network is nitro, if so we call generateL2NitroNetwork to update the contract addresses from the delayed inbox.
https://github.com/OffchainLabs/arbitrum-sdk/blob/1a66fc9e0e4d8658d6ba1d1903409cfd5d579ef5/src/lib/utils/migration_types.ts#L180-L183

However in `L2ToL1Message` constructor we never call `isNitroL1` or `isNitroL2` and hence the network is not generated
https://github.com/OffchainLabs/arbitrum-sdk/blob/1a66fc9e0e4d8658d6ba1d1903409cfd5d579ef5/src/lib/message/L2ToL1Message.ts#L192-L194

yahgwai:
Closes #132

### DIFF
--- a/src/lib/message/L2ToL1Message.ts
+++ b/src/lib/message/L2ToL1Message.ts
@@ -35,7 +35,6 @@ import {
   convertL2ToL1Status,
   IL2ToL1MessageReader,
   IL2ToL1MessageWriter,
-  isNitroL1,
   isNitroL2,
   MessageBatchProofInfo,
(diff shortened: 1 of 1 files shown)


fredlacs (approved):
LGTM

PR #154, opened 2022-08-23
feat: Classic withdrawals in nitro-next

yahgwai:


scripts/genAbi.js
```diff
@@ -42,6 +43,8 @@ async function main() {
   const allFiles = glob(cwd, [
     `${peripheralsPath}/build/contracts/!(build-info)/**/+([a-zA-Z0-9_]).json`,
     `${nitroPath}/build/contracts/!(build-info)/**/+([a-zA-Z0-9_]).json`,
+    // we have a hardcoded abi for the old outbox
+    `./src/lib/dataEntities/OldOutbox.json`,
```
fredlacs, line 47: I believe we want to handle 2 diff "old outboxes"
[arb classic Outbox](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-eth/contracts/bridge/Outbox.sol) and [arb classic OldOutbox](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-eth/contracts/bridge/Old_Outbox/OldOutbox.sol) / outbox entries

src/lib/dataEntities/networks.ts
```diff
@@ -64,6 +64,7 @@ export interface EthBridge {
   sequencerInbox: string
   outbox: string
   rollup: string
+  classicOutbox: string
```
fredlacs, line 67: would be good to make `classicOutbox` optional instead of using `0x0000000000000000000000000000000000000000`

src/lib/message/L2ToL1Message.ts
```diff
@@ -117,94 +49,104 @@ export enum L2ToL1MessageStatus {
...
+          hash,
+          indexInBatch
+        ),
+        nitro.L2ToL1MessageNitro.getL2ToL1Events(
+          l2Provider,
+          filter,
+          position,
+          destination,
+          hash
+        ),
```
fredlacs, line 115: in a future PR we could prob make this more efficient by reducing the number of log queries

src/lib/message/L2ToL1MessageClassic.ts
```diff
@@ -0,0 +1,383 @@
...
+          indexInBatch
+        )
+      : new L2ToL1MessageReaderClassic(
+          l1SignerOrProvider,
+          batchNumber,
+          indexInBatch
+        )
+  }
+
+  public static async getL2ToL1MessageLogs(
```
fredlacs, line 140: we call the similar function in nitro `getL2ToL1Events`, would be good to use the same func names

fredlacs (changes requested):

package.json
```diff
@@ -42,7 +42,8 @@
     "setStandard": "ts-node scripts/setStandardGateways.ts",
     "setCustom": "ts-node scripts/setArbCustomGateways.ts",
     "cancelRetryable": "ts-node scripts/cancelRetryable.ts",
-    "bridgeStandardToken": "ts-node scripts/deployStandard.ts"
+    "bridgeStandardToken": "ts-node scripts/deployStandard.ts",
+    "tt": "typechain --target ethers-v5 --out-dir face './src/lib/message/OutOutbox.json'"
```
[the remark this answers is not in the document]
yahgwai: woops, good catch

scripts/genAbi.js
```diff
@@ -12,6 +12,7 @@ async function main() {
 
   const nitroPath = getPackagePath('@arbitrum/nitro-contracts')
   const peripheralsPath = getPackagePath('arb-bridge-peripherals')
+  const bridgeEthPath = getPackagePath('arb-bridge-eth')
```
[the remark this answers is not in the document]
yahgwai: :+1:

src/lib/dataEntities/networks.ts
yahgwai: updated

src/lib/message/L2ToL1MessageClassic.ts
yahgwai: agreed, updated

scripts/genAbi.js
yahgwai: Added support for both

src/lib/message/L2ToL1Message.ts
yahgwai: yeh, my intuition for a standard index is that this would actually make the query slower as it would introduce OR terms, that standard indexs struggle with. Although ethereum indexes are nothing like standard ones, so I dont know if that applies here. 

One simple thing we can defo do is look at the from/to block and compare to the nitro activation block to adjust the ranges, whilst this will still be two network calls it should be less work for the node. Added that here.

src/lib/message/L2ToL1Message.ts
```diff
@@ -117,94 +50,134 @@ export enum L2ToL1MessageStatus {
...
+    const stringOrMax = (blockTag: BlockTag, nitroGenesisBlock: number) =>
+      typeof blockTag === 'string'
+        ? blockTag
+        : Math.max(blockTag, nitroGenesisBlock)
+
+    // only fetch nitro events after the genesis block
+    const classicFilter = {
+      fromBlock: stringOrMin(filter.fromBlock, nitroGenesisBlock),
+      toBlock: stringOrMin(filter.toBlock, nitroGenesisBlock),
+    }
```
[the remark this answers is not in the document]
fredlacs: would be good to add a test case for this

fredlacs (changes requested):

src/lib/message/L2ToL1Message.ts
```diff
@@ -117,94 +50,134 @@ export enum L2ToL1MessageStatus {
...
-        ArbSys__factory,
-        t => t.filters.L2ToL1Tx(null, destination, hash, position),
-        { ...filter, address: ARB_SYS_ADDRESS }
+    hash?: BigNumber,
+    indexInBatch?: BigNumber
+  ): Promise<(L2ToL1TransactionEvent & { transactionHash: string })[]> {
+    const l2Network = await getL2Network(l2Provider)
+    const nitroGenesisBlock = l2Network.nitroGenesisBlock || 0
+
+    const stringOrMin = (blockTag: BlockTag, nitroGenesisBlock: number) =>
```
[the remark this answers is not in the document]
yahgwai: added a lint rule

src/lib/dataEntities/networks.ts
```diff
@@ -32,6 +32,7 @@ export interface L2Network extends Network {
   isArbitrum: true
   confirmPeriodBlocks: number
   retryableLifetimeSeconds: number
+  nitroGenesisBlock?: number
```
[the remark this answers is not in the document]
yahgwai: made mandatory and added zeros

src/lib/message/L2ToL1Message.ts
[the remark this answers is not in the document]
yahgwai: updated for block tag cases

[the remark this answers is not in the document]
yahgwai: added tests

PR #163, opened 2022-08-31
fix: Also remove timeout defaults for eth deposit messages

yahgwai:


src/lib/dataEntities/networks.ts
```diff
@@ -302,6 +302,34 @@ export const getL2Network = (
...
+    case 42161:
+      return 888000
+    // nova
+    case 42170:
+      return 888000
+    // rinkeby - soon to be decomissioned
+    case 421611:
+      return 9000000
+    default:
+      throw new ArbSdkError(`Unexpected chain id: ${chainId}.`)
```
gzeoneth, line 329: return a default value instead of throw? will fail with local nitro chain id 412346

yahgwai: Moved the deposit timeout onto the network definition, so it will always be there

gzeoneth (approved):
LGTM

PR #166, opened 2022-09-07
Remove `explorerUrl` from networks objects

fionnachan:
> this isn’t used anywhere except the token bridge. we could remove 
> breaking change so should branch off `c-nitro-next`


yahgwai (approved):

gzeoneth:
is there a strong reason we want to remove this? the retryable-dashboard use this too

fionnachan:
decided to keep the `explorerUrl`

PR #173, opened 2022-09-08
ci: add new advisory to audit-ci

fionnachan:


audit-ci.jsonc
```diff
@@ -54,6 +54,10 @@
     // https://github.com/advisories/GHSA-4h98-2769-gh6h
     // OpenZeppelin Contracts vulnerable to ECDSA signature malleability
     // We don’t use signatures for replay protection anywhere
-    "GHSA-4h98-2769-gh6h"
+    "GHSA-4h98-2769-gh6h",
+    // https://github.com/advisories/GHSA-xvch-5gv4-984h
+    // Minimist <=1.2.5 is vulnerable to Prototype Pollution via file index.js, function setKey() (lines 69-95).
+    // GHSA-xvch-5gv4-984h|typedoc-plugin-markdown>handlebars>minimist
```
yahgwai, line 60: When we add an advisory we also add a justification as to why it's ok to ignore

yahgwai:
Lets also add @fredlacs to this PR, it's good to get multiple eyes when adding a new advisory bypass

audit-ci.jsonc
fionnachan: I'd put 
> Minimist is a typedoc plugin's dependency and is never included in the sdk build. Typedoc was only called by the nitro-docs repo to generate sdk docs.

any comments?

yahgwai: Yeh, I think that's good. Maybe mention the words "dev dependency".

yahgwai (approved):

PR #165, opened 2022-09-01
Adding a tx encoder to the SDK

mahsamoosavi:
This is a reference implementation of how to calculate tx hash for Arbitrum Nitro specific EIP-2718 types.

PR #179, opened 2022-09-12
txhasher: adding a txhasher calss for different Arbitrum specific tx types

mahsamoosavi:
The previous [PR](https://github.com/OffchainLabs/arbitrum-sdk/pull/165) had some unrelated changes from another PR.
Had to force push backwards and create a new pull request. I'll be applying all those changes @yahgwai has requested on the previous one!


src/lib/message/txHashEncoder.ts
```diff
@@ -0,0 +1,303 @@
...
+   * @param type //type of the tx
+   * @param l2ChainId
+   * @param fromAddress
+   * @param messageNumber
+   * @param destAddress
+   * @param l1Value
+   * @returns
+   */
+  public static calculateArbitrumDepositTxHash(
+    type: number,
```
mahsamoosavi, line 19: @yahgwai: About [this comment](https://github.com/OffchainLabs/arbitrum-sdk/pull/165#discussion_r964618035), the reason for passing the type is to be able to double check if we're using the right function for encoding. It should be already clear based on the function names but it adds a safety check. 

I can def remove it if you think it's redundant though

mahsamoosavi: nevermind this ^ it's totally restructured now

src/lib/message/ArbTxHasher.ts
```diff
@@ -0,0 +1,301 @@
...
+      case ArbTxType.Contract:
+        return '0x66'
+      case ArbTxType.Retry:
+        return '0x68'
+      case ArbTxType.SubmitRetryable:
+        return '0x69'
+      case ArbTxType.Internal:
+        return '0x6A'
+    }
+  }
```
gzeoneth, line 122: can be 
```
`0x${txType.toString(16)}`
```

src/lib/message/ArbTxHasher.ts
```diff
@@ -0,0 +1,301 @@
...
+   * @returns
+   */
+  public static hash(
+    tx:
+      | DepositTx
+      | UnsignedTx
+      | ContractTx
+      | RetryTx
+      | SubmitRetryableTx
+      | InternalTx
```
gzeoneth, line 136: I think you can have each of the type extend a base type (e.g. with the type field only) so this can be cleaner.

src/lib/message/ArbTxHasher.ts
```diff
@@ -0,0 +1,301 @@
...
+      case ArbTxType.Retry:
+        this.hashRetryTx(tx)
+        break
+      case ArbTxType.SubmitRetryable:
+        this.hashSubmitRetryableTx(tx)
+        break
+      case ArbTxType.Internal:
+        this.hashInternalTx(tx)
+        break
+    }
```
gzeoneth, line 157: Can add 
```
    default:
      const exhaustiveCheck: never = c;
      throw new Error(`Unhandled case: ${exhaustiveCheck}`);
```
as an exhaustive check https://stackoverflow.com/a/58009992

src/lib/message/L1ToL2Message.ts
```diff
@@ -461,17 +485,28 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+      if (confirmations || chosenTimeout) {
+        throw new ArbSdkError(
+          `Timed out waiting to retrieve retryable creation receipt: ${this.retryableCreationId}.`
+        )
+      } else {
+        throw new ArbSdkError(
+          `Retryable creation receipt not found ${this.retryableCreationId}.`
+        )
+      }
+    }
```
gzeoneth, line 509: these unrelated change are still here

gzeoneth (changes requested):
also missing test cases

src/lib/message/ArbTxHasher.ts
yahgwai: Yeah, I like this way of doing it with an UnreachableCaseError: https://stackoverflow.com/a/52913382

src/lib/message/ArbTxHasher.ts
```diff
@@ -0,0 +1,301 @@
...
+      | DepositTx
+      | UnsignedTx
+      | ContractTx
+      | RetryTx
+      | SubmitRetryableTx
+      | InternalTx
+  ) {
+    switch (tx.type) {
+      case ArbTxType.Deposit:
+        this.hashDepositTx(tx)
```
yahgwai, line 140: I think these are missing the return statements - once you add that you also dont need the `break`

src/lib/message/ArbTxHasher.ts
```diff
@@ -0,0 +1,301 @@
...
+        return '0x6A'
+    }
+  }
+
+  /**
+   * Accept the Arbitrum specific tx type and call the right function for creating the custom hash.
+   * @param ArbTxType
+   * @returns
+   */
+  public static hash(
```
yahgwai, line 129: We still need to add tests for this func

src/lib/message/L1ToL2Message.ts
mahsamoosavi: I did reset to c-nitro-stable and force push 
Not sure why they're still there tho

src/lib/message/ArbTxHasher.ts
mahsamoosavi: Adding tests now

package.json
```diff
@@ -84,7 +84,8 @@
     "typedoc": "^0.21.5",
     "typedoc-plugin-markdown": "^3.10.4",
     "typescript": "^4.2.2",
-    "yargs": "^17.3.1"
+    "yargs": "^17.3.1",
+    "ts-essentials": "^9.3.0"
```
fionnachan, line 88: is this package needed?

fionnachan: thanks! can you please undo the yarn.lock changes too?

mahsamoosavi: I did remove the yarn.lock file followed w a yarn install to create a new one
That should do it right?

fionnachan: i think the best way is to do the following:
```bash
git pull
git merge origin/main
```
then resolve the merge conflicts on the files using your IDE
for `yarn.lock`, accept all incoming changes for the whole file then save
stage the changes
complete the merge commit
`git push`

yahgwai:
We should also switch the base branch of this to `main`

PR #183, opened 2022-09-13
chore: Merge from stable

yahgwai:


integration_test/weth.test.ts
```diff
@@ -32,7 +32,7 @@ import { Wallet } from 'ethers'
 import { testSetup } from '../scripts/testSetup'
 import { ERC20__factory } from '../src/lib/abi/factories/ERC20__factory'
 
-describe('WETH', async () => {
+describe.only('WETH', async () => {
```
fionnachan, line 35: Is this intentional?

yahgwai: Nope, good catch, removed!

PR #184, opened 2022-09-13
fix: Parse bytes32 token names

bartlomiej94:


src/lib/utils/multicall.ts
```diff
@@ -329,8 +330,16 @@ export class MultiCaller {
         input.push({
           targetAddr: t,
           encoder: () => erc20Iface.encodeFunctionData('name'),
-          decoder: (returnData: string) =>
-            erc20Iface.decodeFunctionResult('name', returnData)[0] as string,
+          decoder: (returnData: string) => {
```
spsjvc, line 333: Would be cool to write a test case for each of the code paths

spsjvc: Maybe we should do the same thing for token symbol?

src/lib/utils/multicall.ts
```diff
@@ -329,8 +330,16 @@ export class MultiCaller {
...
+            if (arrayify(returnData).length === 32) {
```
yahgwai, line 334: It's unlikely, but I guess there's a chance someone could choose a 32 character name - then we'd try to parse it as bytes. Maybe it's better to check for the maker address instead?

bartlomiej94: is Maker address the only contract we want to cover for this tho? if so that could work, but if there are many contracts using bytes32 couldn't we check if the name string begins with `0x` besides the length check?

fredlacs: instead of checking for the maker address specifically I think we want to detect that its a null terminated string with bytes32 length

yahgwai: I think the null terminated string is probably the safest way - as this should determine that the type was actually bytes32, right @fredlacs? How do you detect a null terminated string returned from ethers?

bartlomiej94: @yahgwai Fred gave me a nice example here: https://github.com/OffchainLabs/token-bridge-contracts/blob/e304fa491f0574c984e80705f9f6d7cf8f0d798e/contracts/tokenbridge/libraries/BytesParser.sol#L43

I will make it similar for our use case

yahgwai: was just talking with @fredlacs and I think the best thing to do is to try to look at the result and see if you can see if a reliable difference between the bytes and string. A string has dynamic length, so it will have some length encoding at the front, in theory this should mean that the result is always actually longer than 32 bytes. But this would be the raw output, i dont know if ethers is decoding to a string before it gets to you. If it is you'll need to figure out a way to make a lower level call that doesn't do the decoding.

bartlomiej94: @yahgwai looks like I'm getting a string when I get the data even if it's given in bytes. I came up with something like this: `utils.isBytesLike(returnData) && utils.hexDataLength(returnData) === 32` - it still doesn't truly differentiate between bytes32 and string but at least this way even if a string is 32 character long then it would also have to be bytelike to get parsed. Let me know your thoughts on this and in the meantime I'll have a look if I can grab bytes32 before ethers parses it to string

yahgwai: yeh, see if you can get it lower in the stack. The best way to do this is to use the `send` (i think it's called that) function on the provider. This function accepts an rpc request object and returns a raw rpc response. You can put some logging somewhere to determine the rpc request you're making atm - or you can do something like `contract.interface.encodeFunctionData(...)` to get the data. Then pass the data as part of an rpc request into `send`

src/lib/utils/multicall.ts
```diff
@@ -329,17 +332,31 @@ export class MultiCaller {
         input.push({
           targetAddr: t,
           encoder: () => erc20Iface.encodeFunctionData('name'),
-          decoder: (returnData: string) =>
-            erc20Iface.decodeFunctionResult('name', returnData)[0] as string,
+          decoder: (returnData: string) => {
+            if (isBytes32(returnData)) {
```
yahgwai, line 336: Lets add a comment here about why we do this (maker doesnt follow the erc20 spec), and include a link to the maker contracts

integration_test/multicall.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      l2Network,
+      latestBlock,
+    }
+  }
+
+  it('returns parsed data from bytes32', async function () {
+    const { l2Provider } = await createProviderMock(421613)
+    const multicaller = await MultiCaller.fromProvider(l2Provider)
+    const [data] = await multicaller.getTokenData(
+      ['0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2'],
```
yahgwai, line 60: Lets comment what the addresses are

integration_test/multicall.test.ts
```diff
@@ -0,0 +1,88 @@
...
+          to: '0x108B25170319f38DbED14cA9716C54E5D1FF4623',
+        }),
+        undefined
+      )
+    ).thenResolve(
+      '0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000204d616b65720000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000204d4b520000000000000000000000000000000000000000000000000000000000'
```
yahgwai, line 32: Can you add a comment of the steps you took to produce these arguments, might be helpful for a future dev when changing these tests

integration_test/multicall.test.ts
```diff
@@ -0,0 +1,88 @@
+'use strict'
+
+import { getL2Network } from '../src/lib/dataEntities/networks'
+import { providers } from 'ethers'
+import { mock, when, anything, instance, deepEqual } from 'ts-mockito'
+import { expect } from 'chai'
+
+import { MultiCaller } from '../src'
+
+describe('Multicall events', () => {
```
yahgwai, line 10: Lets remove the word `events` from the title here - it's a loaded term in the EVM

yahgwai (approved):

PR #186, opened 2022-09-13
fix: Package bumped the nitro dependency and set isnitro checks to true

yahgwai:


src/lib/utils/migration_types.ts
```diff
@@ -129,166 +124,31 @@ export const getOutboxAddr = (
...
-        ...cacheData,
-        l1: {
-          timestamp: Date.now(),
-          value: false,
-        },
-      })
-    }
-  }
-  return cache.getCacheWithDefault(l2ChainId).l1.value
+  // everything is nitro now
```
fredlacs, line 138: does everything go vroom now

fredlacs (approved):
LGTM

PR #189, opened 2022-09-15
fix event filter in getSuccessfulRedeem

Jason-Wanxt:


yahgwai (commented):
We used to have a test that should cover this case. "deposit with low funds, fails first redeem, succeeds seconds" it's currently being skipped. I think we should comment it back in to see if it fails, then apply this change. We started skipping it because nitro has dynamic gas calculations for l1 prices, which caused this test to fail. We should fix the test to account for those.

yahgwai:
Merged in failing tests on https://github.com/OffchainLabs/arbitrum-sdk/pull/192. These should now pass on this branch

yahgwai (approved):

PR #194, opened 2022-09-21
chore: add l2 signed message support in inboxTools

Jason-Wanxt:


src/lib/inbox/inbox.ts
```diff
@@ -40,6 +43,8 @@ type ForceInclusionParams = FetchedEvent<MessageDeliveredEvent> & {
  * Tools for interacting with the inbox and bridge contracts
  */
 export class InboxTools {
+  private readonly L2MSG_signedTx = 4
```
yahgwai, line 46: Create an enum in `message.ts` and add this as the only entry for now

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +306,74 @@ export class InboxTools {
       overrides || {}
     )
   }
+
+  /**
+   * Send l2 signed tx using delayed inox, which won't alias the sender's adddress
+   * It will auto-confirm on l2, if it doesn't confirm in 24 hours, you can force
```
yahgwai, line 312: We usually use the word `include` instead of auto-confirm. Maybe:

```
It will be automatically included by the sequencer on l2, if it isn't included within 24 hours, you can force include it
```

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +306,74 @@ export class InboxTools {
...
+   * include it.
+   * @param message A signed transaction which can be sent directly to network, 
+   *                you can call inboxTools.signL2Message to get.
+   * @returns The l1 delayed inbox's transaction receipt.
```
yahgwai, line 316: This doesnt return the transaction receipt, just the transaction itself i think. They will need to call wait on it after to get the receipt.

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +306,74 @@ export class InboxTools {
...
+   */
+  public async sendL2SignedMessage(
```
yahgwai, line 318: Perhaps this should be called `sendSignedL2Tx`

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +306,74 @@ export class InboxTools {
...
+  }
+
+  /**
+   * Sign a transaction with msg.to, msg.value and msg.data.
+   * You can use this as a helper to call inboxTools.sendL2SignedMessage 
+   * above.
+   * @param message A signed transaction which can be sent directly to network.
+   * @returns The l1 delayed inbox's transaction receipt.
+   */
+  public async signL2Message(
```
yahgwai, line 341: Perhaps this should be called `signL2Tx`

integration_test/sendL2msg.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        l1TransactionReceipt: l1TxReceipt
+    }
+}
+
+describe("Send signedTx to l2 using inbox", () => {
+    it("should confirm the same tx on l2", async () => {
+        const { l2Deployer } = await testSetup()
+        const { signedMsg, l1TransactionReceipt } = await sendSignedTx()
+        const l2Txhash = ethers.utils.parseTransaction(signedMsg).hash!
+        const l2TxReceipt = await l2Deployer.provider!.waitForTransaction(l2Txhash)
```
yahgwai, line 48: We should check that the transaction was actually executed on L2

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +306,74 @@ export class InboxTools {
...
+    tx: {
+      to: string
+      value: BigNumber
+      data: string
+    },
+    l2Signer: Signer
+  ): Promise<string> {
+    const address:string = await l2Signer.getAddress();
+    const nonce = await l2Signer.getTransactionCount()
```
yahgwai, line 350: User should be able to pass in gasPrice and nonce to override

src/lib/inbox/inbox.ts
```diff
@@ -81,6 +86,31 @@ export class InboxTools {
...
+  transactionl2Request: { to: string; data: BytesLike; from: any; value: any },
+  l2Signer: Signer
+  ): Promise<BigNumber>  {
+  const nodeInterface = NodeInterface__factory.connect(
+    NODE_INTERFACE_ADDRESS,
+    l2Signer
+  )
+  const gasComponents = await nodeInterface.callStatic.gasEstimateComponents(
+    transactionl2Request.to,
+    false,
```
yahgwai, line 104: Should allow for contract deploy estimation as well

integration_test/sendL2msg.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+    const signedTx = await inbox.signL2Message(message, l2Deployer)
+    const l1TxReceipt = await inbox.sendL2SignedMessage(signedTx)
+    return {
+        signedMsg: signedTx,
+        l1TransactionReceipt: l1TxReceipt
+    }
+}
+
+describe("Send signedTx to l2 using inbox", () => {
```
yahgwai, line 43: We should also add a test for when the l2gasprice is set to low, and another one for contract deployment

yahgwai (commented):
I think the `sdk` job is failing because of formatting. You can run `yarn format` on your local repo to do the linting.

integration_test/sendL2msg.test.ts
Jason-Wanxt: If l2 gas price is lower than the base fee, network will ignore that tx so it won't record on chain.

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +314,101 @@ export class InboxTools {
...
+    contractCreation: boolean,
+    l2Signer: Signer
+  ): Promise<string> {
+    //check required args
+    if (!((tx.to || contractCreation) && tx.data && tx.value)) {
+      throw new ArbSdkError('Required arg not provided')
+    }
+
+    if (!tx.nonce) {
+      tx.nonce = await l2Signer.getTransactionCount()
```
yahgwai, line 365: It's better not to mutate objects that are passed in to a function. This can result in unexpected things happening for the caller - they may want to use the transaction request elsewhere and be surprised when the properties have changed. Better to create a new object and populate it from the argument

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +314,101 @@ export class InboxTools {
...
+   * @param l2Signer ethers Signer type, used to sign l2 transaction
+   * @returns The l1 delayed inbox's transaction signed data.
+   */
+  public async signL2Tx(
+    tx: TransactionRequest,
+    contractCreation: boolean,
+    l2Signer: Signer
+  ): Promise<string> {
+    //check required args
+    if (!((tx.to || contractCreation) && tx.data && tx.value)) {
```
yahgwai, line 360: You can specify this with the type system by making either `to, data and value` or `contractCreation, data and value` mandatory. The types Required and RequiredPick might be helpful here.

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +314,101 @@ export class InboxTools {
...
+
+    //estimate gas on l2
+    try {
+      tx.gasLimit = await this.estimateGasWithoutL1Part(
+        tx,
+        contractCreation,
+        l2Signer.provider!
+      )
+    } catch (error) {
+      console.log(
```
yahgwai, line 400: We dont console.log in the sdk. If you want to add more info to the error you can use the ArbSdkError which excepts a message and nested error as args

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +314,101 @@ export class InboxTools {
...
+      )
+      throw error
+    }
+
+    // does't need to address when creating contract.
+    if (contractCreation) {
+      delete tx.to
+    }
+
+    const signedTx = await l2Signer.signTransaction(tx)
```
yahgwai, line 411: Minor thing, but you can return here directly instead of assigning.

src/lib/inbox/inbox.ts
```diff
@@ -276,4 +314,101 @@ export class InboxTools {
...
+   * tx.to, tx.data, tx.value must be provided when not contract creation, if
+   * contractCreation is true, no need provide tx.to. tx.gasPrice and tx.nonce
+   * can be overrided.
+   * @param contractCreation this transaction is used to create contract or not.
+   * @param l2Signer ethers Signer type, used to sign l2 transaction
+   * @returns The l1 delayed inbox's transaction signed data.
+   */
+  public async signL2Tx(
+    tx: TransactionRequest,
+    contractCreation: boolean,
```
yahgwai, line 356: We could remove this property and just check to see if `to` is empty, 0x, or the zero address. This is what ethersjs does so it would be good to be consistent with that.

src/lib/inbox/inbox.ts
```diff
@@ -81,6 +93,32 @@ export class InboxTools {
     )
   }
 
+  /**
+   * We should use nodeInterface to get the gas estimate is because we
+   * are making a delayed inbox message which doesn't need l1 calldata
+   * gas fee part.
+   */
+  private async estimateGasWithoutL1Part(
```
yahgwai, line 101: It would probably be more general to return the full result of calling `gasEstimateComponents` then do the subtraction where needed. You could also return the full result and populate an extra property on it `gasEstimateForL2`?

integration_test/sendL2msg.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    signedMsg: signedTx,
+    l1TransactionReceipt: await l1Tx?.wait(),
+  }
+}
+
+describe('Send signedTx to l2 using inbox', () => {
+  it('can deploy contract', async () => {
+    const { l2Deployer } = await testSetup()
+
+    let Greeter = new ethers.ContractFactory(greeter.abi, greeter.bytecode)
```
yahgwai, line 50: Can use const and put the `connect` on the same line as the creation

integration_test/sendL2msg.test.ts
```diff
@@ -0,0 +1,91 @@
...
+import { BigNumber, ethers, providers, Wallet } from 'ethers'
+import { InboxTools } from '../src/lib/inbox/inbox'
+import { getL2Network } from '../src/lib/dataEntities/networks'
+import { testSetup } from '../scripts/testSetup'
+import { greeter } from './helper/greeter'
+import { expect } from 'chai'
+import { isBytes } from '@ethersproject/bytes'
+
+const sendSignedTx = async (contractCreation: boolean, info?: any) => {
+  const { l1Deployer, l2Deployer } = await testSetup()
```
yahgwai, line 29: Better to pass these in rather than call `testSetup` again. Should probably only call `testSetup` once per test as it creates some random addresses.

yahgwai (approved):

yahgwai (commented):
Some minor changes

integration_test/sendL2msg.test.ts
yahgwai: I think that would be a decent test. You could have a test that does
1. Send a tx with gas price too low
2. Send a tx with gas price high enough
3. Wait until tx2 is mined
4. Check that tx1 was never executed on the L2 chain by looking for it's transaction hash

src/lib/inbox/inbox.ts
Jason-Wanxt: return `tx` directly? But this function name is `signL2Tx`

src/lib/inbox/inbox.ts
```diff
@@ -17,25 +17,49 @@
...
 }
 
+type GasComponentsWithL2Part = {
+  gasEstimate: BigNumber
+  gasEstimateForL1: BigNumber
+  baseFee: BigNumber
+  l1BaseFeeEstimate: BigNumber
+  gasEstimateForL2: BigNumber
+}
+type requiredTransactionRequestType = RequiredPick<
```
yahgwai, line 59: We usually capitalise types, so here we'd have `RequiredTransactionRequestType`

yahgwai (approved):

PR #198, opened 2022-09-23
fix: check inbox of bridgeMessageEvent

gzeoneth:
`getL1ToL2Messages` did not check if the bridgeMessageEvent is from the inbox matching the supplied l2provider. 
https://github.com/OffchainLabs/arbitrum-sdk/blob/f6f4c5862baaa6c7537ece45e26d0d7d10d2ff3c/src/lib/message/L1Transaction.ts#L215-L221

If the same L1 tx send to both arb1 and nova some invalid retryable will show up. i.e.

1. nova data with nova chainid
2. arb1 data with nova chainid (invalid)
3. arb1 data with arb1 chainid
4. nova data with arb1 chainid (invalid)

e.g.
https://retryable-dashboard.arbitrum.io/?t=0xc69036ab70acdd288f6db6aff67b6ac1a3f92cfc855d93b5629e1a0ee8f28cbc

PR #200, opened 2022-09-23
fix: Cherry pick - check inbox of bridgeMessageEvent (#198)

yahgwai:
Cherry pick back to stable

* fix: check inbox of bridge message event

* fix: checkRetryableStatus to loop through each msg

* fix: compare lower case


fredlacs (approved):
LGTM

PR #204, opened 2022-09-25
Export Address

DZGoldman:
for access to the aliasing utils :)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -58,4 +58,4 @@ export {
   RetryableDataTools,
 } from './lib/dataEntities/retryableData'
 
-export { Address } from './lib/dataEntities/Address'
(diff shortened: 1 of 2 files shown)


gzeoneth (commented):
I think you can also import by
```
import { Address } from "@arbitrum/sdk/dist/lib/dataEntities/address"
```
without the export in index.ts

src/index.ts
```diff
@@ -57,3 +57,5 @@ export {
   RetryableData,
   RetryableDataTools,
 } from './lib/dataEntities/retryableData'
+
+export { Address } from './lib/dataEntities/Address'
```
yahgwai, line 61: looks like Address shouldn't be capitalised here

yahgwai (approved):

PR #205, opened 2022-09-26
add missing param to natspec

DZGoldman:


yahgwai:
Lets instead add this to `c-nitro-next`

PR #206, opened 2022-09-26
adding EthDepositStatus

GreatSoshiant:
In v3.0.0-rc.1, we used L1ToL2MessageStatus for ETH deposits which doesn't makes sense for "NOT_YET_CREATED" status. In this PR we added EthDepositStatus that represents status for ETH deposits.

### DIFF
--- a/src/lib/message/L1ToL2Message.ts
+++ b/src/lib/message/L1ToL2Message.ts
@@ -68,13 +68,13 @@ export enum L1ToL2MessageStatus {
 
 export enum EthDepositStatus {
   /**
-   * Ethers are not deposited on L2 yet
+   * ETH is not deposited on L2 yet
    */
-   NOT_YET_DEPOSITED_ON_L2 = 1,
+   PENDING= 1,
   /**
-   * Ethers are deposited successfully on L2
+   * ETH is deposited successfully on L2
    */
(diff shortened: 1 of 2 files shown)


src/lib/message/L1ToL2Message.ts
```diff
@@ -66,6 +66,17 @@ export enum L1ToL2MessageStatus {
   EXPIRED = 5,
 }
 
+export enum EthDepositStatus {
+  /**
+   * Ethers are not deposited on L2 yet
+   */
+   NOT_YET_DEPOSITED_ON_L2 = 1,
```
yahgwai, line 73: Maybe lets simplify these status names to "PENDING" and "DEPOSITED".

Also, should probably say "ETH" or "Eth" instead of "Ethers". I know how common it is to use ethers, but that's also the name of a js package, so probably best to avoid here.

yahgwai:
Looks like you need to run `yarn format` locally, then commit and push. That will get the linting checks to pass

src/lib/message/L1ToL2Message.ts
GreatSoshiant: completely makes sense Chris! thanks!

yahgwai (approved):

PR #209, opened 2022-09-30
feat: Added 'from' address to eth and erc20 withdrawal requests

yahgwai:
We do this so that the request returned can be passed directly into estimateGas


fionnachan (approved):
strictly based on the PR's description of its purpose, the changes look good to me

bartlomiej94 (approved):

PR #212, opened 2022-10-03
chore: Port tests from stable

yahgwai:


integration_test/standarderc20.test.ts
```diff
@@ -206,7 +230,7 @@ describe('standard ERC20', () => {
       l2TokenAddr
     )
     // 3 deposits above - increase this number if more deposit tests added
```
fionnachan, line 232: there are 5 now :D 
need to update this comment next time

PR #221, opened 2022-10-31
feat: support old message delivered events

bartlomiej94:
after v3 upgrade pre-nitro message don’t work anymore, so e.g. tx 0xc80e0c4844bb502ed7d7e2db6f9e6b2d52e3d25688de216394f0227fc5dc814e will fail if you try to call getL1ToL2Messages  on its receipt. Couple of points to mention:

- the code for calculating different values for v1 receipts was taken from sdk v1
- before the flow jumps to v1 calculation it checks for whether the current block is pre-nitro (is classic)
- currently working on improving types for the method returning v3 and classic messages

If you're interested in testing these changes locally.

_Prior to all steps below you'll have to set up the nitro node. Follow `Run integration tests` in https://github.com/OffchainLabs/arbitrum-sdk to set it up._

How to reproduce the issue on the main branch:

- create a new file somewhere, e.g. test.ts
- instead of hard-coding the entire txn like in the test, use the following code:

  `const provider = new JsonRpcProvider(YOUR_INFURA_KEY)`
`const arbProvider = new JsonRpcProvider('https://arb1.arbitrum.io/rpc')`
`const txRec = await provider.getTransactionReceipt('0xc80e0c4844bb502ed7d7e2db6f9e6b2d52e3d25688de216394f0227fc5dc814e')`
  `const l1rec = new L1TransactionReceipt(txRec)`
  `const msg = await l1rec.getL1ToL2Messages(arbProvider)`

- run the file: `ts-node ./test.ts`  - you might need to adjust the path to wherever you've put the file
  
  You should see an error saying that messages count doesn't match (or something along these lines)


src/lib/message/L1Transaction.ts
```diff
@@ -196,12 +245,24 @@ export class L1TransactionReceipt implements TransactionReceipt {
   ): Promise<L1ToL2MessageReaderOrWriter<T>[]>
   public async getL1ToL2Messages<T extends SignerOrProvider>(
     l2SignerOrProvider: T
-  ): Promise<L1ToL2MessageReader[] | L1ToL2MessageWriter[]> {
+  ): Promise<
+    L1ToL2MessageReader[] | L1ToL2MessageWriter[] | ClassicL1ToL2MessageReader
```
bartlomiej94, line 249: My concern about adding another type here is potential typescript errors on FE. No errors in SDK related to this.

bartlomiej94:
Creating a draft PR, the solution already works. However still looking into L2toL1Message if things need to be changed there as well

src/lib/message/L1Transaction.ts
bartlomiej94: Added tests, now trying to make it better so that Classic type is not returned together with nitro types, could cause some errors. 

My thinking is maybe we should structure classic data to have similar output to nitro?

bartlomiej94:
@spsjvc @yahgwai thinking of some additional improvements but whatever I've got so far can be reviewed

src/lib/message/L1Transaction.ts
```diff
@@ -58,6 +67,46 @@ export type L1EthDepositTransaction =
...
   L1ContractTransaction<L1ContractCallTransactionReceipt>
 
+class ClassicL1ToL2MessageReader implements ClassicL1ToL2Message {
+  public readonly messageNumber: BigNumber
+  public readonly retryableCreationId: string
+  public readonly autoRedeemId: string
+  public readonly l2TxHash: string
+  public readonly l2Provider: Provider
+
+  constructor(props: {
```
chrstph-dvx, line 77: Is the missing call to `super` intentional? Do we want to extends from `ClassicL1toL2Message`?

bartlomiej94: This class implements ClassicL1toL2Message, doesn't extend it

integration_test/l1toL2MessageEvents.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+describe('L1toL2Message events', () => {
+  it('does call for nitro events', async () => {
+    const testState = {
+      ...(await testSetup()),
+      l1CustomToken: {} as any,
+    }
+
+    await fundL1(testState.l1Signer)
+    await fundL2(testState.l2Signer)
```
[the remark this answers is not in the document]
bartlomiej94: good idea, let me try that instead

src/lib/message/L1Transaction.ts
```diff
@@ -193,15 +242,31 @@ export class L1TransactionReceipt implements TransactionReceipt {
    */
   public async getL1ToL2Messages<T extends SignerOrProvider>(
     l2SignerOrProvider: T
-  ): Promise<L1ToL2MessageReaderOrWriter<T>[]>
+  ): Promise<L1ToL2MessageReaderOrWriter<T>[] | ClassicL1ToL2MessageReader[]>
   public async getL1ToL2Messages<T extends SignerOrProvider>(
     l2SignerOrProvider: T
-  ): Promise<L1ToL2MessageReader[] | L1ToL2MessageWriter[]> {
+  ): Promise<
+    L1ToL2MessageReader[] | L1ToL2MessageWriter[] | ClassicL1ToL2MessageReader[]
```
bartlomiej94, line 249: @yahgwai @spsjvc to avoid having different data type returned from the function we could either:

- pass version prop to the function ('classic' | 'nitro') and based on that we could return correct types. However this way a user will have to provide the 'classic' flag to get the classic data - they will have to figure out themselves whether the txn is classic or nitro

or:

- structure classic data to have the same keys as nitro, so this way ts won't give warnings/errors when trying to access properties

LMK

yahgwai: I think you could do this by extending the L1ToL2MessageReader, and creating a function on the base class for calculating the retryableCreationId and then overriding it in a ClassicL1ToL2MessageReader. Eg.

```
export abstract class L1ToL2Message {

 protected calculateRetryableId(
    chainId: number,
    sender: string,
    messageNumber: BigNumber,
    l1BaseFee: BigNumber,
    messageData: RetryableMessageParams
  ) {
    return L1ToL2Message.calculateSubmitRetryableId(
      chainId,
      sender,
      messageNumber,
      l1BaseFee,
      messageData.destAddress,
      messageData.l2CallValue,
      messageData.l1Value,
      messageData.maxSubmissionFee,
      messageData.excessFeeRefundAddress,
      messageData.callValueRefundAddress,
      messageData.gasLimit,
      messageData.maxFeePerGas,
      messageData.data
    )
  }

  protected constructor(
    public readonly chainId: number,
    public readonly sender: string,
    public readonly messageNumber: BigNumber,
    public readonly l1BaseFee: BigNumber,
    public readonly messageData: RetryableMessageParams
  ) {
    this.retryableCreationId = this.calculateRetryableId(
      chainId,
      sender,
      messageNumber,
      l1BaseFee,
      messageData
    )
  }
}

export class ClassicL1ToL2MessageReader2 extends L1ToL2MessageReader {
  constructor(
    l2Provider: Provider,
    chainId: number,
    sender: string,
    messageNumber: BigNumber,
    l1BaseFee: BigNumber,
    messageData: RetryableMessageParams
  ) {
    super(l2Provider, chainId, sender, messageNumber, l1BaseFee, messageData)
  }

  protected override calculateRetryableId(
    chainId: number,
    sender: string,
    messageNumber: BigNumber,
    l1BaseFee: BigNumber,
    messageData: RetryableMessageParams
  ): string {
    const bitFlip = (num: BigNumber) => num.or(BigNumber.from(1).shl(255))
    return keccak256(
      concat([
        zeroPad(BigNumber.from(chainId).toHexString(), 32),
        zeroPad(bitFlip(this.messageNumber).toHexString(), 32),
      ])
    )
  }
}

```

For the auto redeem id and l2 transaction hash i think you may be able to override other functions too, but I havent looked into it.

src/lib/message/L1Transaction.ts
```diff
@@ -350,7 +415,16 @@ export class L1ContractCallTransactionReceipt extends L1TransactionReceipt {
       message: L1ToL2MessageReaderOrWriter<T>
     } & L1ToL2MessageWaitResult
   > {
-    const message = (await this.getL1ToL2Messages(l2SignerOrProvider))[0]
+    const network = await getL2Network(l2SignerOrProvider)
+
+    if (this.blockNumber < network.nitroGenesisBlock) {
+      throw new ArbSdkError('Unexpected classic transaction.')
```
yahgwai, line 421: I think we can make the `waitForStatus` work instead of throwing an error here

bartlomiej94:
@yahgwai Could you please give it another review? Audit fails here so we need to merge [this PR](https://github.com/OffchainLabs/arbitrum-sdk/pull/232) first

bartlomiej94:
> @yahgwai Could you please give it another review? Audit fails here so we need to merge [this PR](https://github.com/OffchainLabs/arbitrum-sdk/pull/232) first

Fixed the audit error. Can review this one

src/lib/message/L1ToL2Message.ts
```diff
@@ -93,6 +93,9 @@ export abstract class L1ToL2Message {
    * creation of the retryable ticket.
    */
   public readonly retryableCreationId: string
+  // autoRedeemId & l2TxHash used for classic events
+  public readonly autoRedeemId: string | undefined
```
yahgwai, line 97: I dont think these should be on the base class, since they are relevant to all types of L1ToL2Message

src/lib/message/L1Transaction.ts
```diff
@@ -200,8 +240,36 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
+      return [
+        new ClassicL1ToL2MessageReader(
+          provider,
+          BigNumber.from(chainID).toNumber(),
+          '0x',
+          messageNum,
+          BigNumber.from(0),
+          // No message params required for classic events
+          // Populate with empty data
+          {
```
yahgwai, line 257: I think you could remove this from the constructor args, and just set up an empty messageData in the constructor itself. You can do the same for other empty args, like the base fee and sender.

src/lib/message/L1Transaction.ts
```diff
@@ -350,7 +418,16 @@ export class L1ContractCallTransactionReceipt extends L1TransactionReceipt {
       message: L1ToL2MessageReaderOrWriter<T>
     } & L1ToL2MessageWaitResult
   > {
-    const message = (await this.getL1ToL2Messages(l2SignerOrProvider))[0]
+    // const network = await getL2Network(l2SignerOrProvider)
```
yahgwai, line 421: Comments to be removed

src/lib/message/L1ToL2Message.ts
bartlomiej94: In that case I guess we wouldn't be able to read `autoRedeemId ` and `l2TxHash` when having the message object returned? TypeScript will give us errors because we have no way of knowing if tx is classic beforehand (no block number when making the request).

yahgwai: We dont appear to use these in the codebase, should they be used in `waitForStatus`? Since autoredeemid is only useful if you know if the message is classic or not the user should be checking whether the value is null, if they do this they could also cast to the classic message. This approach means the sdk doesnt support classic messages very nicely, but it still makes it possible to use the sdk for this if you know what you're doing. We can chat more about it in catchup?

src/lib/dataEntities/networks.ts
```diff
@@ -284,6 +288,7 @@ export const l2Networks: L2Networks = {
       l2WethGateway: '0x7626841cB6113412F9c88D3ADC720C9FAC88D9eD',
     },
     nitroGenesisBlock: 0,
+    firstNitroBlock: 0,
```
bartlomiej94, line 291: Should this be provided? Otherwise the assumption is only nitro messages are supported.

tests/unit/l1toL2MessageEvents.test.ts
```diff
@@ -0,0 +1,322 @@
...
+      byzantium: true,
+      type: 2,
+      status: 1,
+    }
+
+    const arbProvider = new JsonRpcProvider('https://arb1.arbitrum.io/rpc')
+    const l1TxnReceipt = new L1TransactionReceipt(receipt)
+
+    let txReceipt
+    try {
```
bartlomiej94, line 156: `.to.throw` doesn't work on Promises. This is the solution I could come up with, I'm open to suggestions.

scripts/testSetup.ts
```diff
@@ -115,6 +115,7 @@ export const getCustomNetworks = async (
     partnerChainID: l1NetworkInfo.chainId,
     retryableLifetimeSeconds: 7 * 24 * 60 * 60,
     nitroGenesisBlock: 0,
+    firstNitroBlock: 0,
```
bartlomiej94, line 118: Do we need to set this? If we leave it at 0 it will treat all the txs hitting this and calling message events as nitro

yahgwai: Yeh, I think it's fine to test on nitro

src/lib/dataEntities/networks.ts
yahgwai: Nova was only deployed with nitro, didnt exist in classic times

tests/unit/l1toL2MessageEvents.test.ts
yahgwai: Yeh, i just do it this way too. There some async support for expect, but you need to return it at the end of the func so it doesnt work in lots of cases.

yahgwai (approved):

PR #222, opened 2022-11-01
refactor: cleaner types by inference for `getL1Network` and `getL2Network`

fionnachan:


src/lib/dataEntities/networks.ts
```diff
@@ -288,41 +288,53 @@ export const l2Networks: L2Networks = {
...
 }
-export const getL2Network = (
-  signerOrProviderOrChainID: SignerOrProvider | number
-): Promise<L2Network> => {
-  return getNetwork(signerOrProviderOrChainID, 2) as Promise<L2Network>
+
+type ReturnedNetwork<
+  T extends Network,
+  K extends SignerOrProvider | number
+> = K extends SignerOrProvider ? Promise<T> : T
```
spsjvc, line 312: Should we maybe make these generics a bit more explicit / verbose?

```suggestion
type NetworkOrPromiseOfNetwork<
  TNetwork extends Network,
  TSignerOrProviderOrNumber extends SignerOrProvider | number
> = TSignerOrProviderOrNumber extends SignerOrProvider ? Promise<TNetwork> : TNetwork
```

(comment applies to all functions)

spsjvc (commented):
Works well! Added a comment around making the generics types a bit more easier to understand.

chrstph-dvx (approved):

integration_test/networks.test.ts
```diff
@@ -0,0 +1,485 @@
...
+  L1Network,
+  l1Networks,
+  L2Network,
+  l2Networks,
+} from '../src/lib/dataEntities/networks'
+import { SEVEN_DAYS_IN_SECONDS } from '../src/lib/dataEntities/constants'
+
+/**
+ * This is a unit test rather than an integration test.
+ * TODO -> We'll move it out of integration_test/ when we restructure the tests
```
yahgwai, line 36: I've generally tried to avoid leaving any TODO in the SDK and instead track them via tickets. I find TODOs in the code get forgotten about. Happy to chat more about this though

src/lib/dataEntities/networks.ts
```diff
@@ -288,41 +288,65 @@ export const l2Networks: L2Networks = {
...
+  TNetwork extends Network,
+  TSignerOrProviderOrNumber extends SignerOrProvider | number
+>(
+  signerOrProviderOrChainID: TSignerOrProviderOrNumber,
+  networks: NetworkList<TNetwork>
+): Promise<TNetwork> | TNetwork {
+  if (typeof signerOrProviderOrChainID === 'number') {
+    return getNetworkByChainID(signerOrProviderOrChainID, networks)
+  }
+  return getChainIDfromSignerOrProvider(signerOrProviderOrChainID).then(
```
yahgwai, line 333: Should always `await` before returning, even though there's a slight performance hit to doing this. See: https://github.com/goldbergyoni/nodebestpractices/issues/737

For similar stack trace reasons I prefer not to use `then`, in this case if an error was thrown in getNetworkByChainID you wouldnt get a good stack trace even if the return had been awaited.

src/lib/dataEntities/networks.ts
```diff
@@ -288,41 +288,65 @@ export const l2Networks: L2Networks = {
...
+    chainID => getNetworkByChainID(chainID, networks)
+  )
 }
 
+export const getL1Network = <
+  TSignerOrProviderOrNumber extends SignerOrProvider | number
+>(
+  signerOrProviderOrChainID: TSignerOrProviderOrNumber
+) => getNetwork(signerOrProviderOrChainID, l1Networks)
```
yahgwai, line 342: technically should probably return await here too, but probably doesnt matter if a bit of this stack trace gets skipped.

integration_test/networks.test.ts
```diff
@@ -0,0 +1,485 @@
...
+      // Chain ID 420 is Optimism Goerli
+      expect(() => {
+        getL2Network(420)
+      }).to.throw('Unrecognized network 420.')
+    })
+  })
+
+  context('addCustomNetwork', () => {
+    it('should add the custom L1 and L2 networks to the l1Networks and l2Networks objects respectively', () => {
+      const customL1Network: L1Network = {
```
yahgwai, line 155: We repeat these network configs many times in this file, we could have a function that creates these configs and just call it each time - will be less verbose.

src/lib/dataEntities/networks.ts
fionnachan: was trying to get it not return a Promise for number type arg
let me look into this....

fionnachan:
closing this as the change doesn't seem to improve the concerned code

PR #213, opened 2022-10-04
ci: migrate to github actions

fionnachan:
Cherry picked from #172 and #173

PR #223, opened 2022-11-02
Updated version of nitro in ci to v2.0.8

yahgwai:


fionnachan:
need to get minimist into the advisories list otherwise the audit job will fail
waiting for #213? or we can do a separate PR just for the advisories change

fionnachan:
superseded by #228

PR #226, opened 2022-11-25
Genesis l1 block

Jason-Wanxt:
Get the first transaction on l2's block.number value as it `genesisl1Block` value


src/lib/dataEntities/networks.ts
```diff
@@ -38,6 +38,7 @@ export interface L2Network extends Network {
    * How long to wait (ms) for a deposit to arrive on l2 before timing out a request
    */
   depositTimeout: number
+  genesisL1Block: number
```
yahgwai, line 41: Is this real genesis or nitro genesis? Can you include a comment to explain which it is? Ie either block number 0, or first nitro blocl

Jason-Wanxt: The first transaction's `block.number` return on each l2.

yahgwai (approved):

PR #229, opened 2022-12-07
chore: bump ethers version

bartlomiej94:


fionnachan (commented):
i think we'll have to bump our sdk version and do a release stating it updating ethers from 5.1 to 5.7

package.json
```diff
@@ -48,7 +48,7 @@
     "@ethersproject/address": "^5.0.8",
     "@ethersproject/bignumber": "^5.1.1",
     "@ethersproject/bytes": "^5.0.8",
-    "ethers": "^5.1.0"
+    "ethers": "^5.7.0"
```
chrstph-dvx, line 51: Can we check what would be the minimal version needed for beamer bridge, and add it as a `peerDependency`?

bartlomiej94: Their [frontend repo](https://github.com/beamer-bridge/beamer/blob/main/frontend/package.json#L29) shows version 5.1.1.

What was the context of correlating it to the beamer bridge? Were they the ones having issues with versioning?

spsjvc:
Marked this as draft for now until we discuss the best approach.

Issue #219, opened 2022-10-24
Add ability to deposit to specify a different address when depositing eth


yahgwai:
Currently it's only possible to deposit to yourself but we could use retryable tickets to deposit to a different address.

The annoying thing about this is that it would mean deposits to a different address use the retryable ticket flow, and so a different type system

PR #238, opened 2023-02-06
feat: add method for depositing eth to a different address

TucksonDev:
New `ethBridger.depositTo` method that allows a user to deposit ETH to a different address on L2 by using a Retryable Ticket.
There are different ways to implement the function, but I've prioritized having a simpler function body and just using the parameters to create the retryable ticket (instead of having a more similar function signature to that of `deposit()` and a slightly more complex body).

Added also a test to `eth.test.ts`.

Closes https://github.com/OffchainLabs/arbitrum-sdk/issues/219.

### DIFF
--- a/src/lib/assetBridger/ethBridger.ts
+++ b/src/lib/assetBridger/ethBridger.ts
@@ -208,6 +208,7 @@ export class EthBridger extends AssetBridger<
       ...params,
       to: params.destinationAddress,
       l2CallValue: params.amount,
+      callValueRefundAddress: params.destinationAddress,
       data: '0x',
     }
 
--- a/src/lib/assetBridger/ethBridger.ts
+++ b/src/lib/assetBridger/ethBridger.ts
@@ -27,12 +27,15 @@ import { ARB_SYS_ADDRESS } from '../dataEntities/constants'
 import { AssetBridger } from './assetBridger'
 import {
   L1EthDepositTransaction,
+  L1ContractTransaction,
   L1TransactionReceipt,
 } from '../message/L1Transaction'
 import {
   L2ContractTransaction,
   L2TransactionReceipt,
 } from '../message/L2Transaction'
+import { L1ToL2MessageCreator } from '../message/L1ToL2MessageCreator'
+import { GasOverrides } from '../message/L1ToL2MessageGasEstimator'
 import {
   isL1ToL2TransactionRequest,
   isL2ToL1TransactionRequest,
@@ -76,6 +79,10 @@ export type EthDepositParams = {
    * Transaction overrides
    */
   overrides?: PayableOverrides
+}
+
+export type EthDepositToParams = EthDepositParams & {
+  to: string
 }
 
 export type L1ToL2TxReqAndSigner = L1ToL2TransactionRequest & {
@@ -192,6 +199,37 @@ export class EthBridger extends AssetBridger<
         return BigNumber.from(130000)
       },
     }
+  }
+
+  /**
+   * Deposit ETH from L1 onto a different L2 address
+   * @param params
+   * @returns
+   */
+  public async depositTo(
+    params: EthDepositToParams,
+    l2Provider: Provider,
+    gasOverrides?: GasOverrides
+  ): Promise<L1ContractTransaction> {
+    await this.checkL1Network(params.l1Signer)
+    const from = await params.l1Signer.getAddress()
+
+    const retryableTicketParams = {
+      from: from,
+      to: params.to,
+      l2CallValue: params.amount,
+      callValueRefundAddress: params.to,
+      excessFeeRefundAddress: from,
+      data: '0x',
+    }
+
+    const l1ToL2MessageCreator = new L1ToL2MessageCreator(params.l1Signer)
+    const tx = await l1ToL2MessageCreator.createRetryableTicket(
+      retryableTicketParams,
+      l2Provider,
+      gasOverrides
+    )
+    return tx
   }
 
   /**
--- a/src/lib/assetBridger/ethBridger.ts
+++ b/src/lib/assetBridger/ethBridger.ts
@@ -27,7 +27,7 @@ import { ARB_SYS_ADDRESS } from '../dataEntities/constants'
 import { AssetBridger } from './assetBridger'
 import {
   L1EthDepositTransaction,
-  L1ContractTransaction,
+  L1ContractCallTransaction,
   L1TransactionReceipt,
 } from '../message/L1Transaction'
 import {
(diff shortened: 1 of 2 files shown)


src/lib/assetBridger/ethBridger.ts
```diff
@@ -194,6 +201,37 @@ export class EthBridger extends AssetBridger<
     }
   }
 
+  /**
+   * Deposit ETH from L1 onto a different L2 address
+   * @param params
+   * @returns
+   */
+  public async depositTo(
```
spsjvc, line 209: Can we make this method signature be more closely in line with [Erc20Bridger.deposit](https://github.com/OffchainLabs/arbitrum-sdk/blob/main/src/lib/assetBridger/erc20Bridger.ts#L90), so it has `l2Provider` and `retryableGasOverrides` as part of the `params` object?

src/lib/assetBridger/ethBridger.ts
```diff
@@ -194,6 +201,37 @@ export class EthBridger extends AssetBridger<
...
+    params: EthDepositToParams,
+    l2Provider: Provider,
+    gasOverrides?: GasOverrides
+  ): Promise<L1ContractTransaction> {
```
spsjvc, line 213: I think we want to return `L1ContractCallTransaction` here so we can do `.waitForL2` like in [Erc20Bridger.deposit](https://github.com/OffchainLabs/arbitrum-sdk/blob/main/src/lib/assetBridger/erc20Bridger.ts#L624)

src/lib/assetBridger/ethBridger.ts
spsjvc: Can we also add support for specifying a transaction request instead of params?

spsjvc (commented):
Looking good! 👍 Added a couple of suggestions

src/lib/assetBridger/ethBridger.ts
TucksonDev: Yes to both. I'll add my thoughts:
- I chose to have the signature be closer to EthBridger.deposit instead of Erc20.deposit (I figured it would make sense being in the same file), but it's true that with all the extra parameters it ends up being quite different anyway. I'll change it.
- About the support for specifying a transaction request, I wanted to avoid calling `getDepositRequest` to standarize the request object (because that's already done by `createRetryableTicket`), but I think I can come up with a way to support it and keep the function simple.

src/lib/assetBridger/ethBridger.ts
TucksonDev: Oh, that's right! Yes, definitely. But I'm actually thinking we could change the return type for the `L1ToL2MessageCreator.createRetryableTicket()` . The function hasn't been used until now, so we won't be breaking anything, and it also makes sense to have `.waitForL2` there (plus, I won't need to convert the types in `depositTo()`)
What do you think?

spsjvc: It might not be used internally, but it's possible that it's used externally by someone who installed the package, so changing the return type would be a breaking change. :slightly_smiling_face:

TucksonDev: Oh, I meant that `L1ToL2MessageCreator.createRetryableTicket()` was broken until yesterday ([fix merged here](https://github.com/OffchainLabs/arbitrum-sdk/pull/237)), so it couldn't be used in any way ( unless I'm missing something! ).

src/lib/assetBridger/ethBridger.ts
TucksonDev: We might need the `l2Provider` as a separate parameter. We need it for `L1ToL2MessageCreator.createRetryableTicket()` and is not part of the regular transaction request type (`L1ToL2TxReqAndSigner`, `L1ToL2TxReqAndSignerProvider `and others).
As an alternative, I can also create a new type for the transaction request, extending from an existing one and add that property.
I'm leaning towards adding it as a separate parameter on the function, but I can go either way. What do you think?

(We shouldn't have that problem with GasOverrides as `L1ToL2TransactionRequest` includes those properties)

src/lib/assetBridger/ethBridger.ts
```diff
@@ -164,6 +178,45 @@ export class EthBridger extends AssetBridger<
...
+  /**
+   * Deposit ETH from L1 onto a different L2 address
+   * @param params
+   * @returns
+   */
+  public async depositTo(
+    params: EthDepositToParams | L1ToL2TxReqAndSigner,
+    l2Provider: Provider
+  ): Promise<L1ContractCallTransaction> {
+    await this.checkL1Network(params.l1Signer)
```
yahgwai, line 190: Should we also check the l2 provider?

src/lib/assetBridger/ethBridger.ts
```diff
@@ -164,6 +178,45 @@ export class EthBridger extends AssetBridger<
     return L1TransactionReceipt.monkeyPatchEthDepositWait(tx)
   }
 
+  /**
+   * Deposit ETH from L1 onto a different L2 address
+   * @param params
+   * @returns
+   */
+  public async depositTo(
```
yahgwai, line 186: For each of the functions that send a transaction we currently offer a `getXRequest` variant of the function, this allows people to create the tx and send it later. We then call this `getXRequest` function from with the version of the func that actually sends the transaction - see ethBridger.deposit / ethBridger.getDepositRequest. We should add a `getDepositToRequest` func and use it this way - you can internally use getTicketCreationRequest

src/lib/assetBridger/ethBridger.ts
TucksonDev: Yes, definitely a good idea to avoid weird behaviours. It is used while creating the retryable ticket to get the Inbox contract address.

src/lib/assetBridger/ethBridger.ts
TucksonDev: Oh, good idea. I'll try to add a getDepositToRequest.

src/lib/assetBridger/ethBridger.ts
TucksonDev: https://github.com/OffchainLabs/arbitrum-sdk/pull/238/files#diff-a977cd005aca51be6f05bc7e1c7c1bf6d734b62b2c45c84b05e2eb0c3c3c6fffR233
I ended up adding the `l2Provider` to the `L1ToL2TxReqAndSigner` type in the function signature, so we would only need one parameter for the method. What do you think?

yahgwai (commented):
Only one comment about the usage of getDepositToRequest

yahgwai (approved):

PR #239, opened 2023-02-07
feat: add `isClassic` to `L1TransactionReceipt`

bartlomiej94:


src/lib/message/L1Transaction.ts
```diff
@@ -98,6 +98,18 @@ export class L1TransactionReceipt implements TransactionReceipt {
     this.status = tx.status
   }
 
+  /**
+   * Check if is a classic transaction
+   * @param l2SignerOrProvider
+   */
+  public async isClassic<T extends SignerOrProvider>(
```
yahgwai, line 105: We should call this function inside `getL1ToL2MessagesClassic` and `getL1ToL2Messages`

yahgwai (approved):

PR #240, opened 2023-02-09
feat: add status to classic messages

bartlomiej94:
I think that's all that needs to be returned but lmk if I missed something


src/lib/message/L1ToL2Message.ts
```diff
@@ -553,6 +554,42 @@ export class L1ToL2MessageReaderClassic {
...
+
+    if (!isDefined(creationReceipt)) {
+      return L1ToL2MessageStatus.NOT_YET_CREATED
+    }
+
+    if (creationReceipt.status === 0) {
+      return L1ToL2MessageStatus.CREATION_FAILED
+    }
+
+    return L1ToL2MessageStatus.REDEEMED
```
yahgwai, line 591: Do we need to discern between expired and redeemed? We derive the l2txHash from the retryablecreationid:
```
private static calculateL2DerivedHash(
    retryableCreationId: string,
    l2TxnType: L2TxnType
  ): string {
    return keccak256(
      concat([
        zeroPad(retryableCreationId, 32),
        zeroPad(BigNumber.from(l2TxnType).toHexString(), 32),
      ])
    )
  }
```
then check if that tx was successful:
```
    if (l2TxReceipt && l2TxReceipt.status === 1) {
      return L1ToL2MessageStatus.REDEEMED
    }
```
otherwise the ticket would be EXPIRED.

bartlomiej94: @yahgwai Is `l2TxnType` taken from `creationReceipt.type`? I'm using it to call `calculateL2DerivedHash`, and then running `l2Provider.getTransactionReceipt(
      l2DerivedHash
    )` but it's returning null

src/lib/message/L1ToL2Message.ts
```diff
@@ -553,6 +554,42 @@ export class L1ToL2MessageReaderClassic {
...
+        this.retryableCreationId,
+        confirmations,
+        timeout
+      )
+    }
+
+    return this.retryableCreationReceipt || null
+  }
+
+  public async getMessageStatus(): Promise<L1ToL2MessageStatus> {
```
spsjvc, line 580: Change to `status` to match nitro

yahgwai (approved):

PR #257, opened 2023-03-03
Add api to detect funds in aliased address on L2

TucksonDev:
Some notes:

- After considering several options I decided to add it to the EthBridger as we are dealing with the result of operations performed by this bridge, but I'm open to change it to another class/lib.
- The function only returns a boolean. Another option would be to return the balance, but then the comparison would have to be done in the client side. (This is related to the next point)
- When transferring funds, there's always some left over that stays in the wallet. If a user had funds stuck in their aliased address, when moving them out, there would be some ETH left. Thus, I've added an optional minBalance parameter that defaults to 0.0005 ETH so we check for balances greater than that. Does that make sense?


spsjvc:
Closing this as it's available through https://retryable-tx-panel.arbitrum.io/recover-funds

PR #258, opened 2023-03-10
Add function to add default local network configuration

TucksonDev:


src/lib/dataEntities/networks.ts
```diff
@@ -317,6 +317,72 @@ export const addCustomNetwork = ({
   }
 }
 
+/**
+ * This network configuration is meant to be used for development using the default
+ * smartcontracts deployed on a nitro local node
```
spsjvc, line 322:
```suggestion
 * Registers a custom network that matches the one created by a Nitro local node. Useful in development.
 *
 * @see {@link https://github.com/OffchainLabs/nitro}
```

spsjvc (approved):

PR #260, opened 2023-03-10
feat: add function to obtain eth bridge contracts

TucksonDev:
First attempt to add these functions (tested on local with several generated networks).

Some notes:

- I've added the functions to the `EthBridger` and `ERC20Bridger`. Please let me know if you think they'd make more sense somewhere else
- I've created a new interface for the return value of the `ERC20Bridger.getGatewayInformation()` . That interface looks similar to TokenBridge from `dataEntities/network.ts`, but as I've added two new fields (l1Gateway and l2Gateway) I didn't want to touch too much the one in the network file.
- In that interface, `l1Gateway` and `l2Gateway` correspond to the information of the gateway passed as parameter, which could be a custom gateway. Not sure about the naming, but please let me know of any good ideas you might have.
- The logic of the functions should be simple enough, but I might have missed a shortcut somewhere. Please, let me know if that's the case.
- I've also fixed one logged variable in `deployBridge.ts`


src/lib/assetBridger/ethBridger.ts
```diff
@@ -141,6 +142,36 @@ export class EthBridger extends AssetBridger<
...
+    await this.checkL1Network(l1Provider)
+
+    const rollup = RollupAdminLogic__factory.connect(
+      rollupContractAddress,
+      l1Provider
+    )
+    const bridgeContractAddress = await rollup.bridge()
+    const inboxContractAddress = await rollup.inbox()
+    const outboxContractAddress = await rollup.outbox()
+    const sequencerInboxContractAddress = await rollup.sequencerInbox()
```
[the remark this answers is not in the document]
gzeoneth: I will add that you might not always want to parallelize the RPC calls as it might hit rate limit, I think it is safe to do so here tho.

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -191,6 +219,76 @@ export class Erc20Bridger extends AssetBridger<
...
+
+    try {
+      const l1Gateway = L1ArbitrumGateway__factory.connect(
+        gatewayAddress,
+        l1Provider
+      )
+
+      gatewayInformation.l1Gateway = gatewayAddress
+      gatewayInformation.l2Gateway = await l1Gateway.counterpartGateway()
+      gatewayInformation.l1GatewayRouter = await l1Gateway.router()
```
gzeoneth, line 255: `gateway.router()` is not part of the ITokenGateway interface, for example DAI and LPT gateway use `.l1router()`
in generally I don't think we need to return the router here, if needed they can be retrieved from the network object, you are also putting trust into the gateway to return the correct router address

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -191,6 +219,76 @@ export class Erc20Bridger extends AssetBridger<
...
+    } catch (error) {
+      // It's an L2 Gateway
```
gzeoneth, line 257: you should not try catch to detect L1 or L2, they might have the same address deployed on both side

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -191,6 +219,76 @@ export class Erc20Bridger extends AssetBridger<
...
+      const l2Gateway = L2ArbitrumGateway__factory.connect(
+        gatewayAddress,
+        l2Provider
+      )
+      gatewayInformation.l2Gateway = gatewayAddress
+      gatewayInformation.l1Gateway = await l2Gateway.counterpartGateway()
+      gatewayInformation.l2GatewayRouter = await l2Gateway.router()
```
gzeoneth, line 264: can be retrieved from the network object

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -191,6 +219,76 @@ export class Erc20Bridger extends AssetBridger<
...
+        await l2GatewayRouter.counterpartGateway()
+    }
+
+    const l1GatewayRouter = L1GatewayRouter__factory.connect(
+      gatewayInformation.l1GatewayRouter,
+      l1Provider
+    )
+    gatewayInformation.l2GatewayRouter =
+      await l1GatewayRouter.counterpartGateway()
+    gatewayInformation.l1ERC20Gateway = await l1GatewayRouter.defaultGateway()
```
gzeoneth, line 280: can be retrieved from the network object

gzeoneth (changes requested):

src/lib/assetBridger/ethBridger.ts
```diff
@@ -141,6 +142,36 @@ export class EthBridger extends AssetBridger<
...
+  public async getBridgeInformation(
+    rollupContractAddress: string,
+    l1Provider: Provider
+  ): Promise<EthBridge> {
+    await this.checkL1Network(l1Provider)
+
+    const rollup = RollupAdminLogic__factory.connect(
+      rollupContractAddress,
+      l1Provider
+    )
```
gzeoneth, line 160: you don't need this, there are `getL1Network` and `getL2Network` that return the same thing
https://github.com/OffchainLabs/arbitrum-sdk/blob/5f622e10f5cb871edb90c3c2bb6e39b1e2d0cf0f/src/lib/dataEntities/networks.ts#L274-L283

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -191,6 +219,76 @@ export class Erc20Bridger extends AssetBridger<
...
+      gatewayInformation.l2Gateway = await l1Gateway.counterpartGateway()
+      gatewayInformation.l1GatewayRouter = await l1Gateway.router()
+    } catch (error) {
+      // It's an L2 Gateway
+      const l2Gateway = L2ArbitrumGateway__factory.connect(
+        gatewayAddress,
+        l2Provider
+      )
+      gatewayInformation.l2Gateway = gatewayAddress
+      gatewayInformation.l1Gateway = await l2Gateway.counterpartGateway()
```
gzeoneth, line 263: you should not trust the l2Gateway without validating this piece of info with the l1Gateway

gzeoneth:
so I think the only thing you might want to add here is `getL1CounterpartGateway` and `getL2CounterpartGateway`. Also note that `counterpartGateway` is also not part of the token gateway interface, and you should validate the data the gateway return against the other side.

TucksonDev:
Changed logic so only the EthBridge information is retrieved from now.
Moved it to the dataEntities/networks file as having it in the EthBridger file requires an L2Network to be set.

src/lib/dataEntities/networks.ts
```diff
@@ -282,6 +283,42 @@ export const getL2Network = (
...
+  ])
+
+  return {
+    bridge: bridgeContractAddress,
+    inbox: inboxContractAddress,
+    sequencerInbox: sequencerInboxContractAddress,
+    outbox: outboxContractAddress,
+    rollup: rollupContractAddress,
+  }
+}
```
spsjvc, line 320: Couple of things:
* I think we don't need `await` before the contract calls, as they will return promises, so we get an array of promises that we can await with `Promise.all`
* Nitpick: If we name the properties we destructure from the array the same as the object properties, we can use the shorthand object thing

```suggestion
  const [
    bridge,
    inbox,
    outbox,
    sequencerInbox,
  ] = await Promise.all([
    rollup.bridge(),
    rollup.inbox(),
    rollup.outbox(),
    rollup.sequencerInbox(),
  ])

  return {
    bridge,
    inbox,
    sequencerInbox,
    outbox,
    rollup: rollupContractAddress,
  }
}
```

tests/unit/ethBridgeAddresses.test.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '../../src/lib/dataEntities/networks'
+import dotenv from 'dotenv'
+dotenv.config()
+
+/**
+ * Tests the getEthBridgeInformation function against the information
+ * of Arbitrum One network (42161)
+ */
+describe('Obtain deployed bridge addresses', () => {
+  it('obtains deployed ETH Bridge addresses', async () => {
```
spsjvc, line 16: I'd move this test to integration tests since we're actually hitting the network and we're not mocking anything

gzeoneth (approved):
LGTM

PR #263, opened 2023-03-22
add ethbridge proxy admin addresses to networks config

DZGoldman:


TucksonDev:
A couple of things:

- If we change the interface, we need to add the proxyAdmin to the `defaultLocalL2Network` object in `network.ts` (I don't know what's the address though, or how to get it!).
- We also need to return the address of the proxyAdmin when generating the local network, in `scripts/testSetup.ts`, in the `getCustomNetworks()` function. We might obtain it from the `docker exec nitro_sequencer_1 cat /config/deployment.json` command that is run when trying to get the local config.
- Can you run the linter to add those missing commas and remove the warnings?

PR #266, opened 2023-04-20
docs: autodoc generated doc

fionnachan:
I ran autodoc on node v20.0.0 so probably had some issues with that
It was erroring out at the folder step because it was looping

normal output is:
image


fionnachan:
closing this as it's difficult to craft a prompt that fits all files to generate meaningful docs

PR #290, opened 2023-06-05
feat: add `ARB_OWNER_PUBLIC` to constants

shotaronowhere:
I ran into a dead end trying to query the arbOwner pre-compile before realizing I need to call the arbOwnerPublic precompile at a different address. Including the constant for the arbOwnerPublic is not only a better developer experience, but can help clarify that developers are probably interested in the arbOwnerPublic precompile, not the arbOwner precompile


gzeoneth (approved):
LGTM

PR #294, opened 2023-06-12
Adds `ViemSigner` compatibility layer  

douglance:
- AssetBridgers can be instantiated with a provider from ethers v5, v6, web3, or viem.
  - **Question**: Do we want to remove the support for ethers v6 and web3?
- Signing can be done with ethers v5 or viem signers.
  - consuming packages must instantiate a `ViemSigner` using `createViemSigner` before passing in the signer to the sdk.


src/lib/utils/universal/providerTransforms.ts
fionnachan, line 1: we should unit test all the functions in this file

tests/integration/universal/universalSigner.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  },
+}
+const arbRpcUrl = config.arbUrl
+
+export const ethLocal = {
+  ...mainnet,
+  id: 1337,
+  rpcUrls: {
+    default: {
+      http: ['http://localhost:8545'],
```
fionnachan, line 32: should use env var value here too

scripts/testSetup.ts
```diff
@@ -236,11 +251,22 @@ export const testSetup = async (): Promise<{
...
+      id: 1337,
+      rpcUrls: {
+        default: {
+          http: [config.ethUrl],
+        },
+        public: {
+          http: [config.ethUrl],
+        },
+      },
+    },
```
fionnachan, line 266: why wouldn't you use `ethLocal` here?

src/lib/utils/universal/signerTransforms.ts
```diff
@@ -127,21 +89,20 @@ class ViemSigner extends Signer {
       // https://viem.sh/docs/actions/wallet/prepareTransactionRequest.html#account-hoisting
       requestData
     )
-    const serializedTransaction = await this.walletClient.signTransaction(
-      request
-    )
-    const hash = await this.publicClient.sendRawTransaction({
-      serializedTransaction,
+    const hash = await this.walletClient.sendTransaction({
+      ...request,
```
fionnachan, line 93: Is this to create a copy and make sure `request` is not changed?
It is viem's responsibility to ensure the function doesn't create side effects to `request`, and by the look of its code we should pass `request` directly
https://github.com/wevm/viem/blob/main/src/actions/wallet/sendTransaction.ts#L121

unless we are passing additional params than `request`, I don't see why we should spread it

tests/unit/providerTransforms.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      expect(isEthersV5JsonRpcProvider(provider)).to.be.false
+    })
+
+    it('should return false for a Viem PublicClient', () => {
+      const provider = createPublicClient({
+        transport: http('http://localhost:8545'),
+      })
+      expect(isEthersV5JsonRpcProvider(provider)).to.be.false
+    })
+  })
```
fionnachan, line 45: this set of description doesn't match with what's in the tests

the tests are creating **different providers** and testing each of them against **whether they are ethers v5 provider**

whereas the description says
`here's an ethers v5 provider, it should return an ethers v5 provider`,
`here's an ethers v5 provider, (?) should correctly identify it's a v5 provider`,
`here's an ethers v5 provider, (?) should correctly identify it's not a v6 provider`,
`here's an ethers v5 provider, (?) should correctly identify it's not a web3 provider`

so i'd either change the tests to:

```typescript
describe('Provider Utilities', () => {
  describe('ethers v5', () => {
    it('should return the URL for an actual EthersV5 provider', () => {
      const provider = new JsonRpcProvider('http://localhost:8545')
      const url = getEthersV5Url(provider)
      expect(url).to.be.equal('http://localhost:8545')
    })

    it('should be correctly identified as an EthersV5 JsonRpcProvider', () => {
      const provider = new JsonRpcProvider('http://localhost:8545')
      expect(isEthersV5JsonRpcProvider(provider)).to.be.true
    })

    it('should not be identified as an EthersV6 provider', () => {
      const provider = new JsonRpcProvider('http://localhost:8545')
      expect(isEthersV6JsonRpcProvider(provider)).to.be.false
    })
   ...
  })
```
or
```typescript
describe('Provider Utilities', () => {
  describe('isEthersV5JsonRpcProvider', () => {
    it('should correctly identify an EthersV5 JsonRpcProvider', () => {
      const provider = new JsonRpcProvider('http://localhost:8545')
      const url = getEthersV5Url(provider)
      expect(url).to.be.equal('http://localhost:8545')
      expect(isEthersV5JsonRpcProvider(provider)).to.be.true
    })

    it('should correctly identify an EthersV6 provider is not an EthersV5 provider', () => {
      const provider = new JsonRpcProviderv6('http://localhost:8546')
      expect(isEthersV5JsonRpcProvider(provider)).to.be.false
    })
   ...
  })
```
something along those lines ^
feel free to figure out what's best

PR #300, opened 2023-06-16
Fix retryableExists

DZGoldman:
ensure errors that aren't call reverts (i.e., RPC timeout errors) just throw / don't cause retryableExists to return false (which can lead to `status()` incorrectly reporting "expired"


TucksonDev (approved):
LGTM

PR #311, opened 2023-07-24
refactor `addCustomNetwork` by validating both L1 and L2 networks prior to adding them

bartlomiej94:
We want to allow users to add custom networks via the bridge UI and ideally we'd want the method to validate both networks before pushing the results.

Otherwise, if there is an error in the L2 object, we still store L1, even though the function eventually throws.

### DIFF
--- a/src/lib/dataEntities/networks.ts
+++ b/src/lib/dataEntities/networks.ts
@@ -330,8 +330,6 @@ export const addCustomNetwork = ({
       throw new ArbSdkError(
         `Custom network ${customL1Network.chainID} must have isCustom flag set to true`
       )
-    } else {
-      l1Networks[customL1Network.chainID] = customL1Network
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/lib/dataEntities/networks.ts
```diff
@@ -343,9 +341,13 @@ export const addCustomNetwork = ({
     )
   }
 
-  l2Networks[customL2Network.chainID] = customL2Network
+  let l1PartnerChain = l1Networks[customL2Network.partnerChainID]
+
+  // Case where the added L1 network is a partner of the added L2 network.
+  if (customL2Network.partnerChainID === customL1Network?.chainID) {
```
yahgwai, line 347: I think I dont understand the problem, as I'm not sure I understand what's happening here, should not always be the case that the provided custom network should have matching partner chain ids?

Shouldnt the behaviour be:
1. custom network already exists in the state - therefore dont supply one here, just connect
2. custom network does not already exist - therefore supply one here and add it

bartlomiej94: I was implementing a feature in the bridge to allow users to add custom networks through our UI.

However, because L1 network is added to the list before L2 validation, users could run into an issue where:

1. They add a json config where L1 is valid but L2 has a mistake.
2. Following the current logic, L1 would be added successfully but L2 would throw. Our UI would show an error that adding networks has failed. However, L1 was successful.
3. They fix the json so that L2 is correct, but now they can't use the same json because L1 network is already added.

Overall bad UX. What I did here is I verify both L1 and L2 networks first, and only at the end of the method when we know both are valid, we add them to the list.

bartlomiej94: Going back to the mentioned change, since L1 and L2 networks are added at the end after both are validated, we need to also check either is true:

1. Existing L1 network is a parent to the added L2 network, OR
2. Added L1 network is a parent to the added L2 network.

yahgwai: Would it be better to check that l2 and l1 are valid at the very top of this function, and throw an error if they aren't?

bartlomiej94: I'm closing this PR because the required logic is added in https://github.com/OffchainLabs/arbitrum-sdk/pull/312. We can use a separate method not to overcomplicate existing stuff.

PR #312, opened 2023-08-02
support L2 parent chains

bartlomiej94:


src/lib/inbox/inbox.ts
```diff
@@ -90,7 +90,13 @@ export class InboxTools {
 
     // we take a long average block time of 14s
     // and always move at least 10 blocks
-    const diffBlocks = Math.max(Math.ceil(diff / this.l1Network.blockTime), 10)
+    let diffBlocks = 10
+    if ((this.l1Network as L1Network).blockTime) {
```
bartlomiej94, line 94: `L2Network` does not have `blockTime`. I'm not sure if we need to add `blockTime` to every L2 network that is a parent chain to L3?

src/lib/dataEntities/networks.ts
```diff
@@ -78,7 +80,7 @@ export interface EthBridge {
 }
 
 export interface L1Networks {
-  [id: string]: L1Network
+  [id: string]: L1Network | L2Network
```
bartlomiej94, line 83: Wouldn't this change be considered too breaking?

src/lib/dataEntities/networks.ts
```diff
@@ -28,14 +28,16 @@ export interface L1Network extends Network {
...
   tokenBridge: TokenBridge
   ethBridge: EthBridge
   partnerChainID: number
-  isArbitrum: true
+  isArbitrum: boolean
   confirmPeriodBlocks: number
   retryableLifetimeSeconds: number
   nitroGenesisBlock: number
   nitroGenesisL1Block: number
+  isL3?: boolean
```
spsjvc, line 40: Do we need this property? Can we compute that something is an Orbit chain like this:

```
network.isArbitrum && network.partnerChainIDs.length > 0
```

src/lib/dataEntities/networks.ts
```diff
@@ -28,14 +28,16 @@ export interface L1Network extends Network {
 }
 
 export interface L2Network extends Network {
+  partnerChainIDs: number[]
```
spsjvc, line 31: This will actually be a breaking change, so we should make it optional probably, and let's add jsdoc

We can also then remove the changes of adding this to all the networks

src/lib/dataEntities/networks.ts
```diff
@@ -28,14 +28,16 @@ export interface L1Network extends Network {
...
   tokenBridge: TokenBridge
   ethBridge: EthBridge
   partnerChainID: number
-  isArbitrum: true
+  isArbitrum: boolean
```
spsjvc, line 35: I think we can keep this as `true`, no?

src/lib/dataEntities/networks.ts
spsjvc: Shouldn't be, as it's a union type so a superset of the previous one

src/lib/dataEntities/networks.ts
```diff
@@ -265,11 +289,24 @@ const getNetwork = async (
...
-  } else {
-    throw new ArbSdkError(`Unrecognized network ${chainID}.`)
+  const network = networks[chainID]
+
+  if (network) {
+    if (layer === 2) {
+      return network
+    }
+
+    // Any network that's a parent chain is a valid L1 network
```
spsjvc, line 299:
```suggestion
    // Any network that's a parent chain is valid
```

src/lib/dataEntities/networks.ts
bartlomiej94: Do we want Orbit chains to be flagged as Arbitrum tho?

src/lib/dataEntities/networks.ts
bartlomiej94: we can do that

src/lib/dataEntities/networks.ts
bartlomiej94: What if you try to access properties that are on L1Network but not on L2Network? You'll have to check for undefined first as typescript will complain.

src/lib/dataEntities/networks.ts
bartlomiej94: what if `network` is an `L2Network`, and we want to check if `network` is L3? then we need to do something like:

```
await getL2Network(network.parentChainID).parentChainIDs.length > 0
```

it seems a bit complicated: to figure out if the chain of interest is an L3 we'd have to get its parent chain first.

src/lib/dataEntities/networks.ts
spsjvc: Well they are all Arbitrum chains, so I would say yes. Not sure if that messes up anything, could you check?

src/lib/dataEntities/networks.ts
spsjvc: Let's create a type `ParentChains` that could either be `L1Network` or `L2Network`, and `Chains` that can be `L2Network`.

We should also create separate `getParentChain` and `getChain` instead of editing `getL1Network` and `getL2Network`

spsjvc: We'd also need `parentChains` that mimics `l1Networks` but adds L2s that are parent chains. We also need to use that when instantiating a bridger class

src/lib/dataEntities/networks.ts
```diff
@@ -28,10 +28,14 @@ export interface L1Network extends Network {
 }
```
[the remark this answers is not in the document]
bartlomiej94: It will be left as `false` here because `L1Network` doesn't include Arbitrum chains. If an Arbitrum chain is a parent chain, it will be part of the `ParentChain`.

src/lib/inbox/inbox.ts
gzeoneth: see https://github.com/OffchainLabs/arbitrum-sdk/pull/319 for an attempted fix

src/lib/dataEntities/networks.ts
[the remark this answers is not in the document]
gzeoneth: https://github.com/OffchainLabs/arbitrum-sdk/pull/319 added a util method to detect if a chain is Arbitrum chain runtime

src/lib/inbox/inbox.ts
bartlomiej94: I didn't include this fix because it's out of scope of this PR and the bridge UI doesn't require.
I can create a separate PR to add this.

gvladika (commented):
It would probably be useful to have couple of tests for adding/fetching chains with their parents

src/lib/dataEntities/networks.ts
```diff
@@ -325,7 +368,67 @@ export const getL1Network = (
 export const getL2Network = (
   signerOrProviderOrChainID: SignerOrProvider | number
 ): Promise<L2Network> => {
-  return getNetwork(signerOrProviderOrChainID, 2) as Promise<L2Network>
+  let l2Network
+  try {
+    l2Network = getNetwork(signerOrProviderOrChainID, 2) as Promise<L2Network>
+  } catch {
+    l2Network = getChain(signerOrProviderOrChainID) as Promise<Chain>
```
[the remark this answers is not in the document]
bartlomiej94: They could, but that also means we'd have to change all instances in the sdk where `getL2Network` is used, to make it work with Orbit chains. That is the most minimal solution I could come up with.

src/lib/dataEntities/networks.ts
```diff
@@ -293,6 +309,28 @@ export const l2Networks: L2Networks = {
...
 }
 
+// L2 networks that are a parent chain to Orbit chains.
+function getL2ParentChains(_l2Networks: L2Networks) {
+  const _parentChains: ParentChains = {}
+
+  Object.keys(_l2Networks).map(chainId => {
+    const network = l2Networks[Number(chainId)]
+
+    if (network.partnerChainIDs && network.partnerChainIDs.length > 0) {
```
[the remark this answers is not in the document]
bartlomiej94: It shouldn't be the case because our list only has 'child chains' in `partnerChainIDs`. If L2 network's L1 chain partner is specified in `partnerChainID` which is a number, not an array.

Are you concerned that someone would mistakenly add an L1 chain ID to the L2 partnerChainIDs list?

src/lib/dataEntities/networks.ts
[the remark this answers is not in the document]
bartlomiej94: We wouldn't have to modify it at all if we got read of `Chains`, and used `L2Networks` instead. WDYT?

src/lib/dataEntities/networks.ts
```diff
@@ -28,6 +28,7 @@ export interface L1Network extends Network {
 }
 
 export interface L2Network extends Network {
+  partnerChainIDs?: number[]
```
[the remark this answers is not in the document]
bartlomiej94: Why do we need to do that? L2Network is always Arbitrum

src/lib/dataEntities/networks.ts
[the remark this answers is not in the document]
gvladika: I see, terminology is bit confusing, it would be better if we had `parentChain` and `childChains`. But I understand we need to keep backward compatibility so it's ok to keep as is

src/lib/dataEntities/networks.ts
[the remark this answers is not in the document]
gvladika: got it, then we can keep it like this I think

gvladika (approved):
lgtm

src/lib/dataEntities/networks.ts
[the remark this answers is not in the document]
bartlomiej94: Yes all Orbit chains are Arbitrum

gzeoneth (commented):
Works, but there are some concern about increased code debt (e.g. usage of e.g. `partnerChainIDs` to infer L2 parent chain and newly exported type Chain/ParentChain) before we can refactoring everything from L1/L2 to Parent/Child chain.

bartlomiej94:
Moving to draft as it's in beta, not merging right now

PR #317, opened 2023-08-10
chore: `getBlockRangesForL1Block`

bartlomiej94:


src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -204,16 +204,39 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
     nodeNum: BigNumber,
     l2Provider: Provider
   ): Promise<ArbBlock> {
-    const node = await rollup.getNode(nodeNum)
+    const { createdAtBlock } = await rollup.getNode(nodeNum)
+
+    let createdFromBlock = createdAtBlock
+    let createdToBlock = createdAtBlock
+
+    const isL2Arbitrum = [42161, 421613].includes(
```
bartlomiej94, line 212: Do we need to include Nova?

src/lib/utils/lib.ts
```diff
@@ -52,3 +55,109 @@ export const getTransactionReceipt = async (
...
+
+  // Binary search to find the starting Arbitrum block that corresponds to the L1 block number.
+  async function getL2StartBlock() {
+    let result
+    let start = startArbBlock
+    let end = endArbBlock
+
+    while (start <= end) {
+      // Calculate the midpoint of the current range.
+      const mid = start + Math.floor((end - start) / 2)
```
gzeoneth, line 90: I am a little bit unsure about these casting between BigNumber and number, it should be fine and I used it somewhere else in the sdk too.

src/lib/utils/lib.ts
```diff
@@ -52,3 +55,109 @@ export const getTransactionReceipt = async (
...
+        result = mid
+        start = mid + 1
+      } else if (l1Block < targetL1BlockNumber) {
+        // If the L1 block number is less than the target, adjust the range to the upper half.
+        start = mid + 1
+      } else {
+        // If the L1 block number is greater than the target, adjust the range to the lower half.
+        end = mid - 1
+      }
+    }
```
gzeoneth, line 139: we now have binary search on block in at least 3 different place of the sdk codebase, maybe we can try to refactor it to reduce code duplication?

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -204,16 +204,39 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
...
   ): Promise<ArbBlock> {
-    const node = await rollup.getNode(nodeNum)
+    const { createdAtBlock } = await rollup.getNode(nodeNum)
+
+    let createdFromBlock = createdAtBlock
+    let createdToBlock = createdAtBlock
+
+    const isL2Arbitrum = [42161, 421613, 42170].includes(
+      (await l2Provider.getNetwork()).chainId
+    )
```
gzeoneth, line 214: this can be detected runtime, see https://github.com/OffchainLabs/arbitrum-sdk/pull/319
also we can move this into `getBlockRangesForL1Block `, which is also done in the above PR

src/lib/utils/lib.ts
```diff
@@ -52,3 +55,109 @@ export const getTransactionReceipt = async (
...
+
+    if (typeof result === 'undefined') {
+      throw new Error(`No L2 range found for L1 block: ${targetL1BlockNumber}`)
```
gzeoneth, line 142: this should not throw as this is very likely to happen

we should return `the first L3 block that have a l1 block number >= X` (for getL2StartBlock)
or `the last L3 block that have a l1 block number <= X` (for getL2EndBlock)
we might want to rename the function to better reflect this behavior

then, in `getBlockRangesForL1Block` if there are no such range we can have a defined behavior of retuning a crossing range e.g. [3,2], and the client logic can decide what to do with that information

src/lib/utils/lib.ts
```diff
@@ -52,3 +55,109 @@ export const getTransactionReceipt = async (
...
+    }
+
+    return BigNumber.from(result)
+  }
+
+  // Binary search to find the ending Arbitrum block that corresponds to the L1 block number.
+  async function getL2EndBlock() {
```
gzeoneth, line 116: nice to have these not as nested functions, as these can be useful when only 1 of the boundries is of interest e.g. in
https://github.com/OffchainLabs/arbitrum-sdk/blob/71a3e955b1347a51c9ffd54f9b0beaf9d5b12a49/src/lib/inbox/inbox.ts#L88

gzeoneth (changes requested):

src/lib/utils/lib.ts
```diff
@@ -56,108 +57,125 @@ export const getTransactionReceipt = async (
...
+ * @param {boolean} [allowGreater=false] - Whether to allow the search to go past the specified `forL1Block`.
+ * @param {number|string} [minL2Block='latest'] - The minimum L2 block number to start the search from. Can be a number or `'latest'`. `'latest'` means it will be close to the current block, but not exactly the current block.
+ * @param {number|string} [maxL2Block='latest'] - The maximum L2 block number to end the search at. Can be a `number` or `'latest'`. `'latest'` is the current block.
+ * @returns {Promise<{ l2Block: number | undefined; forL1Block: number | undefined }>} - A Promise that resolves to an object containing the L2 block number and the corresponding L1 block number.
+ */
+export async function getFirstBlockForL1Block({
   provider,
+  forL1Block,
+  allowGreater = false,
+  minL2Block = 'latest',
```
gzeoneth, line 84: the should default to rollup genesis block or 0

src/lib/utils/lib.ts
```diff
@@ -56,108 +57,125 @@ export const getTransactionReceipt = async (
...
   }
 
-  // Binary search to find the starting Arbitrum block that corresponds to the L1 block number.
-  async function getL2StartBlock() {
-    let result
-    let start = startArbBlock
-    let end = endArbBlock
+  if (minL2Block === 'latest') {
+    minL2Block = Math.floor(currentArbBlock * 0.95)
+  }
```
gzeoneth, line 110: should not allow minBlock to be latest

src/lib/utils/lib.ts
```diff
@@ -56,108 +57,125 @@ export const getTransactionReceipt = async (
...
+  return { l2Block: lastValidL2Block, forL1Block: resultForL1Block }
+}
+
+export const getBlockRangesForL1Block = async (props: {
+  forL1Block: number
+  provider: JsonRpcProvider
+}) => {
+  return await Promise.all([
+    getFirstBlockForL1Block({ ...props, allowGreater: false }),
+    getFirstBlockForL1Block({ ...props, allowGreater: true }),
```
gzeoneth, line 179: I haven't really looked at the logic yet, but if implemented as what I expected this should be
```suggestion
    getFirstBlockForL1Block({ props.forL1Block + 1, props.provider, allowGreater: true }) - 1,
```

src/lib/utils/lib.ts
```diff
@@ -52,3 +56,132 @@ export const getTransactionReceipt = async (
...
+  maxL2Block = 'latest',
+}: {
+  provider: JsonRpcProvider
+  forL1Block: number
+  allowGreater?: boolean
+  minL2Block?: number
+  maxL2Block?: number | 'latest'
+}): Promise<number | undefined> {
+  if (!isArbitrumChain(provider)) {
+    throw new Error('Arbitrum provider is required.')
```
gzeoneth, line 94: I suggest we assume the provider is L1 if its not Arbitrum, and return `forL1Block`

tests/unit/utils.test.ts
gzeoneth, line 1: getBlockRangesForL1Block(17926532) -> [121907680, 121907739]
getBlockRangesForL1Block(17926533) -> undefined
getFirstBlockForL1Block(17926532) -> 121907680
getFirstBlockForL1Block(17926533, false) -> undefined
getFirstBlockForL1Block(17926533, true) -> 121907740

also add these here maybe?

gzeoneth (commented):
mostly lgtm, some comments

src/lib/utils/lib.ts
bartlomiej94: can do

tests/unit/utils.test.ts
gzeoneth: hmm, you can supply a minBlock and maxBlock based on the expected answer to improve the speed I think

gzeoneth (approved):
LGTM

PR #343, opened 2023-09-27
perf: cache l2 block range

brtkx:
Implements mutex lock to cache l2BlockRange

### DIFF
--- a/src/lib/message/L2ToL1MessageNitro.ts
+++ b/src/lib/message/L2ToL1MessageNitro.ts
@@ -65,7 +65,7 @@ const ASSERTION_CONFIRMED_PADDING = 20
 const l2BlockRangeCache: { [key in string]: (number | undefined)[] } = {}
 const mutex = new Mutex()
 
-export function getL2BlockRangeCacheKey({
+function getL2BlockRangeCacheKey({
   chainId,
   l1BlockNumber,
 }: {
--- a/src/lib/message/L2ToL1MessageNitro.ts
+++ b/src/lib/message/L2ToL1MessageNitro.ts
@@ -66,32 +66,32 @@ const l2BlockRangeCache: { [key in string]: (number | undefined)[] } = {}
 const mutex = new Mutex()
 
 function getL2BlockRangeCacheKey({
-  chainId,
+  l2ChainId,
   l1BlockNumber,
 }: {
-  chainId: number
+  l2ChainId: number
   l1BlockNumber: number
 }) {
-  return `${chainId}-${l1BlockNumber}`
+  return `${l2ChainId}-${l1BlockNumber}`
 }
 
 function setL2BlockRangeCache(key: string, value: (number | undefined)[]) {
   l2BlockRangeCache[key] = value
 }
 
-async function getL2BlockRangeCache({
+async function getBlockRangesForL1BlockWithCache({
   l1Provider,
   l2Provider,
-  l1BlockNumber,
+  forL1Block,
 }: {
   l1Provider: JsonRpcProvider
   l2Provider: JsonRpcProvider
-  l1BlockNumber: number
+  forL1Block: number
 }) {
-  const { chainId } = await l2Provider.getNetwork()
+  const l2ChainId = (await l2Provider.getNetwork()).chainId
   const key = getL2BlockRangeCacheKey({
-    chainId,
-    l1BlockNumber,
+    l2ChainId,
+    l1BlockNumber: forL1Block,
   })
 
   if (l2BlockRangeCache[key]) {
@@ -109,7 +109,7 @@ async function getL2BlockRangeCache({
 
   try {
     const l2BlockRange = await getBlockRangesForL1Block({
-      forL1Block: l1BlockNumber,
+      forL1Block,
       provider: l1Provider,
     })
     setL2BlockRangeCache(key, l2BlockRange)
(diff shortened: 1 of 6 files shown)


src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -61,6 +61,8 @@ const ASSERTION_CREATED_PADDING = 50
 // expected number of L1 blocks that it takes for a validator to confirm an L1 block after the node deadline is passed
 const ASSERTION_CONFIRMED_PADDING = 20
 
+const l2BlockRangeCache: { [key in number]: number[] } = {}
```
spsjvc, line 64: This needs to also be grouped by parent chain

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -155,6 +157,10 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
     return outbox.callStatic.isSpent(this.event.position)
   }
 
+  public setL2BlockRangeCache(l1Block: number, l2BlockRange: number[]) {
+    l2BlockRangeCache[l1Block] = l2BlockRange
+  }
```
spsjvc, line 162: I would remove this method and just directly update the cache above from the code

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -212,17 +218,24 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
...
         if (!startBlock || !endBlock) {
           throw new Error()
         }
         createdFromBlock = BigNumber.from(startBlock)
         createdToBlock = BigNumber.from(endBlock)
+
+        this.setL2BlockRangeCache(createdAtBlock.toNumber(), [
+          startBlock,
+          endBlock,
+        ])
```
spsjvc, line 238: Maybe we should cache it below the try/catch

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -212,17 +218,24 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
...
-        const l2BlockRange = await getBlockRangesForL1Block({
-          forL1Block: createdAtBlock.toNumber(),
-          provider: this.l1Provider as JsonRpcProvider,
-        })
+        const l2BlockRange =
+          l2BlockRangeCache?.[createdAtBlock.toNumber()] ||
+          (await getBlockRangesForL1Block({
+            forL1Block: createdAtBlock.toNumber(),
+            provider: this.l1Provider as JsonRpcProvider,
+          }))
```
spsjvc, line 226: nit for readability

```suggestion
        const cachedL2BlockRange = l2BlockRangeCache?.[createdAtBlock.toNumber()]
        
        return cachedL2BlockRange ?? (await getBlockRangesForL1Block({
          forL1Block: createdAtBlock.toNumber(),
          provider: this.l1Provider as JsonRpcProvider,
        }))
```

src/lib/message/L2ToL1MessageNitro.ts
spsjvc: My bad, needs to be grouped by chain, cause one parent chain (e.g. Arb Goerli) can have many Orbit chains

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -211,11 +231,21 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
...
+        chainId,
+        l1BlockNumber: createdAtBlock.toNumber(),
+      })
+
       try {
-        const l2BlockRange = await getBlockRangesForL1Block({
-          forL1Block: createdAtBlock.toNumber(),
-          provider: this.l1Provider as JsonRpcProvider,
-        })
+        const cachedL2BlockRange = getL2BlockRangeCache(l2BlockRangeCacheKey)
```
gzeoneth, line 242: we should acquire a lock here on cache miss to block parallel call

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -228,6 +258,10 @@ export class L2ToL1MessageReaderNitro extends L2ToL1MessageNitro {
         createdFromBlock = createdAtBlock
         createdToBlock = createdAtBlock
       }
+
+      if (l2BlockRange) {
+        setL2BlockRangeCache(l2BlockRangeCacheKey, l2BlockRange)
```
gzeoneth, line 263: then release the lock here

gzeoneth (commented):
While caching `(chainId, l1BlockNumber) -> l2BlockRange` can save around 50% of the rpc call for unconfirmed node (since we always call `getBlockFromNodeNum` on latest confirmed), a better way seems to be caching the `(chainId, blockNumber) -> l1BlockNumber` result. 

By using a bitwise binary search, it should be possible to save more than 75% rpc calls for unconfirmed node (50% from the latest confirmed, another 50%+ because of the duplicated binary search in `getBlockRangesForL1Block`) and the fact that those l2 blocks are nearby

A bitwise binary search improve cache hit since you will have the same midpoint assuming the l2 blocks are in nearby range (which is always `>latestConfirmed`).

gzeoneth (approved):
LGTM

src/lib/message/L2ToL1MessageNitro.ts
```diff
@@ -61,6 +62,64 @@ const ASSERTION_CREATED_PADDING = 50
...
+
+async function getL2BlockRangeCache({
+  l1Provider,
+  l2Provider,
+  l1BlockNumber,
+}: {
+  l1Provider: JsonRpcProvider
+  l2Provider: JsonRpcProvider
+  l1BlockNumber: number
+}) {
```
spsjvc, line 90:
```suggestion
async function getBlockRangesForL1BlockWithCache({
  l1Provider,
  l2Provider,
  forL1Block,
}: {
  l1Provider: JsonRpcProvider
  l2Provider: JsonRpcProvider
  forL1Block: number
}) {
```

spsjvc (approved):

PR #346, opened 2023-09-28
Add support for local L3 in `testSetup`

godzillaba:
* replaced `arb-bridge-peripherals` dependency with `@arbitrum/token-bridge-contracts`
* modify `deployBridge.ts` 
    * rename `deployErc20L1` and `deployErc20L2` to `deployErc20Parent` and `deployErc20Child`
    * modify `deployErc20AndInit` to deploy and init "parent" and "child" token bridge contracts to L1, L2, L3
* modify `testSetup.ts`
    * essentially, everything that it did before for L1 <-> L2, it does for L2 <-> L3 too
    * one exception though is that `testSetup` doesn't return `Erc20Bridger`, `EthBridger`, `AdminErc20Bridger`, or `InboxTools` for L3

PR #353, opened 2023-10-02
WIP: L1-L3 Teleporter

godzillaba:


src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+  bridgeToL3: L1ToL2MessageWaitResult
+  /**
+   * Whether the teleportation has completed
+   */
+  completed: boolean
+}
+
+/**
+ * When using the L1Teleporter the second leg is a retryable tx, so this type includes the status of that tx as well as the base type's `l2ForwarderCall`.
+ * This is because the retryable could possibly be frontran and the teleportation will still succeed.
```
DZGoldman, line 176:
```suggestion
 * This is because the retryable could possibly be frontrun and the teleportation will still succeed.
```

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+  /**
+   * Check the signer/provider matches the l2Network, throws if not
+   * @param sop
+   */
+  protected async _checkL2Network(sop: SignerOrProvider): Promise<void> {
+    await SignerProviderUtils.checkNetworkMatches(sop, this.l2Network.chainID)
+  }
+
+  /**
+   * Check the signer/provider matches the l2Network, throws if not
```
DZGoldman, line 290:
```suggestion
   * Check the signer/provider matches the l3Network, throws if not
```

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+      await this.getL2ERC20Address(erc20L1Address, l1Provider),
+      l2Provider,
+      this.l3Network
+    )
+  }
+
+  /**
+   * Given an erc20 address on a parent network, get the corresponding erc20 address on a child network
+   */
+  private _getChildErc20Address(
```
DZGoldman, line 363: code in erc20Bridger.getL2ERC20Address could be reused (not necessary tho)

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+    const parentGatewayRouter = L1GatewayRouter__factory.connect(
+      childNetwork.tokenBridge.l1GatewayRouter,
+      parentProvider
+    )
+
+    return parentGatewayRouter.calculateL2TokenAddress(erc20ParentAddress)
+  }
+
+  /**
+   * Get the address of the L1 <-> L2 gateway on L1 given an L1 token address
```
DZGoldman, line 378:
```suggestion
   * Given an L1 token's address, get the address of the token's L1 <-> L2 gateway on L1 
```

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+    l1Provider: Provider,
+    l2Provider: Provider
+  ): Promise<string> {
+    await this._checkL1Network(l1Provider)
+    await this._checkL2Network(l2Provider)
+
+    const l2Token = await this.getL2ERC20Address(erc20L1Address, l1Provider)
+
+    return await L1GatewayRouter__factory.connect(
+      this.l3Network.tokenBridge.l1GatewayRouter,
```
DZGoldman, line 406: so l1GateRouter is really on L2 here, wondering if a clarifying comment would help? (might just be worse tho lol)

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+
+    if (!log) {
+      throw new ArbSdkError(`Could not find TransferRouted event in tx receipt`)
+    }
+
+    // topic 3 is "_userTo"
+    const bytes32UserTo = log.topics[3]
+
+    // parse address
+    return ethers.utils.getAddress(bytes32UserTo.slice(26))
```
DZGoldman, line 539: I think you can alternatively use 

getAddress(hexDataSlice(bytes32UserTo, 12))

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+   * check the underlying form of the contract bytecode to see if it's an erc20, and doesn't ensure the validity
+   * of any of the underlying functions on that contract.
+   */
+  public getL1TokenContract(l1TokenAddr: string, l1Provider: Provider): ERC20 {
+    return ERC20__factory.connect(l1TokenAddr, l1Provider)
+  }
+
+  /**
+   * Get the L2 token contract at the provided address
+   * Note: This function just returns a typed ethers object for the provided address, it doesnt
```
DZGoldman, line 450:
```suggestion
   * Note: This function just returns a typed ethers object for the provided address, it doesn't
```

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+      erc20L1Address,
+      l1Provider,
+      l2Provider
+    )
+    return gateway === this.l3Network.tokenBridge.l1ERC20Gateway
+  }
+
+  /**
+   * Get the L1 token contract at the provided address
+   * Note: This function just returns a typed ethers object for the provided address, it doesnt
```
DZGoldman, line 440:
```suggestion
   * Note: This function just returns a typed ethers object for the provided address, it doesn't
```

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+  protected async _getDepositStatus(
+    depositTxReceipt: L1ContractCallTransactionReceipt,
+    l2Provider: JsonRpcProvider,
+    l3Provider: Provider
+  ): Promise<BaseErc20DepositStatus> {
+    await this._checkL2Network(l2Provider)
+    await this._checkL3Network(l3Provider)
+
+    const l1l2Messages = await depositTxReceipt.getL1ToL2Messages(l2Provider)
+    const firstLegRedeem = await l1l2Messages[0].getSuccessfulRedeem()
```
DZGoldman, line 556: Maybe handle the multiple messages case? I.e., either have an index param or return an array of statuses

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1166 @@
...
+ * Bridge ERC20 tokens from L1 to L3 using the L1Teleporter contract.
+ */
+export class Erc20L1L3Bridger extends BaseErc20L1L3Bridger {
+  /**
+   * Get a tx request to approve tokens for teleportation.
+   * The tokens will be approved to be spent by the L1Teleporter.
+   */
+  public async getApproveTokenRequest(
+    params: TokenApproveParams
+  ): Promise<Required<Pick<TransactionRequest, 'to' | 'data' | 'value'>>> {
```
DZGoldman, line 704: can value be left out here?

src/lib/assetBridger/l1l3Bridger.ts
DZGoldman: (okay realiziing this gets a little weird since there's 2 retryables; maybe something like "offsetIndex" but idk)

DZGoldman (approved):
small things, looks good overall

src/lib/assetBridger/l1l3Bridger.ts
godzillaba: I'm not sure I understand, to account for someone creating a tx that does other unrelated retryables?

src/lib/assetBridger/l1l3Bridger.ts
godzillaba: agree that's nicer

scripts/genAbi.js
```diff
@@ -32,16 +33,17 @@ async function main() {
     cwd: nitroPath,
   })
 
-  console.log('building peripherals')
-  execSync(`${npmExec} && ${npmExec} run hardhat compile`, {
-    cwd: peripheralsPath,
```
gzeoneth, line 37: I don't think the peripheral build should be removed

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1141 @@
...
+  }
+
+  /**
+   * Get the corresponding L2 token address for the provided L1 token
+   */
+  public getL2ERC20Address(
+    erc20L1Address: string,
+    l1Provider: Provider
+  ): Promise<string> {
+    return new Erc20Bridger(this.l2Network).getL2ERC20Address(erc20L1Address, l1Provider)
```
gzeoneth, line 336: instead of instantiate a Erc20Bridger everytime, maybe we keep a singleton instance?

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1141 @@
...
+
+  /**
+   * Get the corresponding L3 token address for the provided L1 token
+   */
+  public async getL3ERC20Address(
+    erc20L1Address: string,
+    l1Provider: Provider,
+    l2Provider: Provider
+  ): Promise<string> {
+    return new Erc20Bridger(this.l3Network).getL2ERC20Address(
```
gzeoneth, line 347: same here

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1141 @@
...
+    const topic0 =
+      L1GatewayRouter__factory.createInterface().getEventTopic('TransferRouted')
+    const log = depositTxReceipt.logs.find(x => x.topics[0] === topic0)
+
+    if (!log) {
+      throw new ArbSdkError(`Could not find TransferRouted event in tx receipt`)
+    }
+
+    // topic 3 is "_userTo"
+    const bytes32UserTo = log.topics[3]
```
gzeoneth, line 512: can use the `parseTypedLog` util

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1141 @@
...
+    BigNumber.from(30)
+
+  /**
+   * Get a tx request to approve tokens for teleportation. Will approve the token's L1 gateway to spend the provided amount.
+   */
+  public async getApproveTokenRequest(
+    params: TokenApproveParams,
+    l1Provider: Provider
+  ): Promise<Required<Pick<TransactionRequest, 'to' | 'data'>>> {
+    return new Erc20Bridger(this.l2Network).getApproveTokenRequest({
```
gzeoneth, line 828: it seems weird to use the ERC20Bridger class here just for the approval, maybe instead have BaseErc20L1L3Bridger.approveToken with an optional param to allow specifying the approval address

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1141 @@
...
+      ...baseStatus,
+      retryableL2ForwarderCall: secondStepRetryableRedeem,
+    }
+  }
+}
+
+/**
+ * Bridge ERC20 tokens from L1 to L3 directly through the token bridge, using a relayer on L2 to call the L2Forwarder contract.
+ */
+export class RelayedErc20L1L3Bridger extends BaseErc20L1L3Bridger {
```
gzeoneth, line 814: I don't feel like having this class in the sdk, it add complexity to the sdk while the relayer feature is unlikely to be used by default. Would prefer to simply having this option documented but not implemented in the sdk (maybe create a tutorial instead).  One risk being if the deposit method crashed due to whatever reason the relayer info might be lost which will lead to fund loss as mentioned. 
If we were to keep this, an option can be appending the relayer config to the deposit transaction to store that info onchain at the expense of increased calldata cost.

gzeoneth (changes requested):
some change requested, also:
- run `yarn format` to fix lint and format issues
- overall speaking this seems to have a lot of duplicated code, would be nice if we can reduce them by using internal function / optional parameters instead

scripts/genAbi.js
godzillaba: whenever i try to `hardhat compile` in that directory it just says nothing to compile since there's no hardhat config. the npm package also already has the `build` dir

gzeoneth: I see, #346 bumped from classic peripheral to the new token-bridge-contracts repo. That would require some additional hacking since the hardhat config is removed from the token-bridge-contracts release. An example fix is here https://github.com/OffchainLabs/arbitrum-sdk/pull/295

godzillaba: ok cool, I'll do something like that in 346 then

src/lib/assetBridger/l1l3Bridger.ts
godzillaba: i really like the idea of appending the relayer info to calldata. i agree that it's a good bit of extra complexity, but I also think it's valuable to have in the sdk given the cost savings.

appending the calldata could also make the api simpler because you wouldn't really need to return / provide relayerinfo anymore, just pass a tx hash or receipt around.

gzeoneth: if there are no risk of losing fund I think that's fine

godzillaba: 5d7388a994fca9a0dc6aeff1ad70d8d0235a005c

tests/integration/l1l3Bridger.test.ts
```diff
@@ -0,0 +1,556 @@
...
+          },
+          setup.l1Signer,
+          setup.l2Signer.provider!,
+          setup.l3Signer.provider!
+        )
+
+        const depositReceipt = await depositTx.wait()
+
+        // poll status
+        await poll(async () => {
```
yahgwai, line 82: Is it possible to something like below instead of manually polling?
```
const depositTx = await l1l3Bridger.deposit(...);
const depositReceipt = await depositTx.wait()
const l1tol2Message = depositReceipt.getL1ToL2Messages()[0]
const l2TxReceipt = new L1TransactionReceipt(await l1ToL2Message.waitForL2())
const l2tol3Message = depositReceipt.getL1ToL2Messages()[0]
await l2toL3Message.waitForL2()
```

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1187 @@
...
+      await l1Signer.sendTransaction(txRequest)
+    )
+  }
+
+  /**
+   * Get the status of a deposit given an L1 tx receipt. See `Erc20DepositStatus` interface for more info on the return type.
+   *
+   * Note: This function does not verify that the tx is actually a deposit tx.
+   */
+  public async getDepositStatus(
```
yahgwai, line 750: Generally we've put the status methods on the messages themselves. In this case we dont have an l1toL3 message, I'm not sure if it makes sense to have one or not, what do you think?

If we do keep this functions here I feel like they could be simplified by using the existing status functions on the underlying l1tol2Messages

godzillaba:
> so instead of hacking with calldata, I am thinking if we should simply modifier the L1 contract to accept additional parameters and calculate the address (instead of calculating the address offchain)

@gzeoneth If we want to use a relayer to save costs and not touch any additional L1 contracts i think we have to calculate offchain. I assume you're talking about https://github.com/OffchainLabs/arbitrum-sdk/commit/5d7388a994fca9a0dc6aeff1ad70d8d0235a005c?

src/lib/assetBridger/l1l3Bridger.ts
```diff
@@ -0,0 +1,1187 @@
...
+      to: this.teleporterAddresses.l1Teleporter,
+      data: calldata,
+      value: calculatedGasCosts.total,
+    }
+  }
+
+  /**
+   * Deposit tokens to L3 through the L1Teleporter contract.
+   */
+  public async deposit(
```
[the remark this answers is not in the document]
godzillaba: good point, I will add this

src/lib/assetBridger/l1l3Bridger.ts
godzillaba: i'm not sure. i put it here initially because it was easier.

one argument though for keeping it here is that the statuses will be different for each type of L1-L3 transfer:
standard erc20 bridge: 2x L2, 1x L3 initiated by the second L2 message
relayed erc20: 1x L2, 1x L3 initiated by a relayer
ETH: 1x L2, 1x L3 initiated by L2 message

whereas with the existing L1L2 message status functions it's just a single message.

tests/integration/l1l3Bridger.test.ts
godzillaba: i had a function like this before instead of `getDepositStatus` but the full stack team said they wouldn't use it and would instead do polling. I thought it might be best to just take it out

src/lib/assetBridger/l1l3Bridger.ts
yahgwai: yeh, i think it's ok to keep it - unless we make a full l1tol2message class.

However internally i think we should use the status() functions to calculate the return of getDepositStatus

tests/integration/l1l3Bridger.test.ts
yahgwai: Oh, fair enough.

Here though I mean that we could just do the above code instead of polling to make your test simpler.

godzillaba:
moved to #396

PR #371, opened 2023-11-03
fix: default block to 0 if no NodeCreated events found

brtkx:
Fixes an issue with newly deployed Orbit chains when there are no assertions yet. In that case, we return block 0.


gzeoneth (approved):
LGTM

spsjvc (approved):

PR #374, opened 2023-11-15
fix: wait for the receipts of initialising txs

ImJeremyHe:
We are trying to use an external sequencer, and the ordering of these transactions is not guaranteed. I don't know if the order matters here, but we fix the problems by this.


fionnachan:
Hi @ImJeremyHe can you please sign the CLA?

hkalodner:
@cla-bot check

gzeoneth (approved):
LGTM

PR #378, opened 2023-11-18
reduce external API changes for orbit support

douglance:
- combines the `l1Network` & `l2Network` objects into a singlular `Networks` object
- adds queries into the `Networks` object for common situations
- uses queries to populate generated `l1Network` and `l2Network` objects when adding new network
- adds more robustness around typechecking different types of networks
- adds `isOrbit` property when adding a custom orbit chain
- makes the `layer` argument of `getNetwork()` optional.
  - if `layer` is not specified it get network info from any layer, including orbit chains


src/lib/dataEntities/networks.ts
```diff
@@ -48,6 +48,7 @@ export type ParentChain =
   | (L2Network & Required<Pick<L2Network, 'partnerChainIDs'>>)
 
 export type Chain = L2Network
+export type OrbitChain = Chain & { isOrbit: boolean }
```
spsjvc, line 51: Let's remove `Chain` and `ParentChain` entirely. Also, for an `OrbitChain` type, can we make `isOrbit: true` in the type? 
```suggestion
export type OrbitChain = L2Network & { isOrbit: true }
```

src/lib/dataEntities/networks.ts
```diff
@@ -346,31 +350,100 @@ export const l2Networks: L2Networks = {
...
+  return Object.entries(Networks).reduce((acc, [key, value]) => {
+    if (isOrbitChain(value)) {
+      acc[key] = value
+    }
+    return acc
+  }, {} as OrbitChains)
+}
 
-const getNetwork = async (
+export const getParentOfNetwork = (chain: ChainOrNetwork) => {
```
spsjvc, line 417:
```suggestion
export const getParentForNetwork = (chain: ChainOrNetwork) => {
```

src/lib/dataEntities/networks.ts
```diff
@@ -346,31 +350,100 @@ export const l2Networks: L2Networks = {
...
+  const parentChain = Networks[chain.partnerChainID]
+  if (!parentChain || !isParentChain(parentChain)) {
+    throw new ArbSdkError(
+      `ParentChain ${chain.partnerChainID} not recognized for Chain ${chain.chainID}.`
+    )
+  }
+  return parentChain
+}
+
+export const getChildrenOfNetwork = (chain: ChainOrNetwork) => {
```
spsjvc, line 430:
```suggestion
export const getChildrenForNetwork = (chain: ChainOrNetwork) => {
```

PR #379, opened 2023-11-23
feat: add support for L3 chains

spsjvc:
### L2Network

Added the following properties to make `L2Network` more consistent with `L1Network`:
  - `blockTime: number` represents the minimum block time of the chain
    - Added a new constant `ARB_MINIMUM_BLOCK_TIME_IN_SECONDS`
    - In v4, we can think about dropping this property altogether
  - `partnerChainIDs: number[]` represents the ids of the children chains
    - In v4, we can simplify this by just having a `parentChainId` property on the `L2Network` and reconstructing the tree of networks when necessary

The name `L2Network` is currently wrongly named, and it should be changed in a future version to `ArbitrumNetwork` or `ArbitrumChain`, because it can represent an L2 or an L3.
### Bridger, InboxTools

Updated the type of the `l1Network` property from `L1Network` to `L1Network | L2Network` as it's referencing the parent chain which can now be an `L2Network`. The changes mentioned above should make this seamless.


src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
+const isL2Chain = (chain: Chain): chain is L2Network => {
+  return chain && isChildChain(chain) && !isOrbitChain(chain)
+}
+
+const isOrbitChain = (chain: Chain): chain is OrbitChain => {
+  return chain && 'isOrbit' in chain && chain.isOrbit
+}
+
+export const getL1Chains = () => {
+  return Object.entries(Networks).reduce((acc, [key, value]) => {
```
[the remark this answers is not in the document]
brtkx: true, I've misinterpreted the result at first

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
+
+const isL2Chain = (chain: Chain): chain is L2Network => {
+  return chain && isChildChain(chain) && !isOrbitChain(chain)
+}
+
+const isOrbitChain = (chain: Chain): chain is OrbitChain => {
+  return chain && 'isOrbit' in chain && chain.isOrbit
+}
+
+export const getL1Chains = () => {
```
fionnachan, line 371: same comment regarding the name, `getBaseChains`

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
+}
+
+export const getOrbitChains = () => {
+  return Object.entries(Networks).reduce((acc, [key, value]) => {
+    if (isOrbitChain(value)) {
+      acc[key] = value
+    }
+    return acc
+  }, {} as OrbitChains)
+}
```
fionnachan, line 414: I don't know how much a specific result type matter so this might not work
but if possible maybe abstract them out to a common function like this?

```typescript
export const getChains = <T extends Chain>(
  isChainType: (chain: Chain) => chain is T
) => {
  return Object.entries(Networks).reduce((acc, [key, value]) => {
    if (isChainType(value)) {
      acc[key] = value
    }
    return acc
  }, {} as Record<string, T>) // or { [id: string]: T }
}

export const getL1Chains = () => getChains(isL1Chain)
export const getL2Chains = () => getChains(isL2Chain)
export const getParentChains = () => getChains(isParentChain)
export const getChildChains = () => getChains(isChildChain)
export const getOrbitChains = () => getChains(isOrbitChain)
```

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
+
+export const getParentForNetwork = (chain: Chain) => {
```
fionnachan, line 416: we're calling them Chains sometimes and Networks sometimes
do we want to use one name across everything?

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
+  return children
+}
+
+export let l1Networks: L1Networks = getL1Chains()
+export let l2Networks: L2Networks = getL2Chains()
+
+export const getNetwork = async (
   signerOrProviderOrChainID: SignerOrProvider | number,
-  layer: 1 | 2
+  layer?: 1 | 2
```
fionnachan, line 445: why is this changed to optional now?
i see from the tests that if it's not passed it's an orbit chain?
wouldn't it be better if we type it to `1 | 2 | 'orbit'` or sth like that?

tests/unit/network.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const mainnetId = 1
+const arbOneId = 42161
+const mockOrbitChainId = 99999999
+
+describe('Network', () => {
+  const fetchErrorMessage =
+    'Network fetched successfully but the chain ID is invalid.'
+
+  it('Adds a custom Orbit chain', async function () {
```
fionnachan, line 14: nit: across the code base's test files, the `it` test cases starts with lower case

tests/unit/network.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      customL2Network: {
+        // we partially copy Arbitrum One network because we only want to mimic a custom chain
+        ...arbOneNetwork,
+        chainID: mockOrbitChainId,
+        partnerChainID: arbOneId,
+        isArbitrum: true,
+        isCustom: true,
+        isOrbit: true,
+      },
+    })
```
fionnachan, line 27: shouldn't there be a check for `Networks` containing this network after `addCustomNetwork`?

tests/unit/network.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        .be.undefined
+    }
+  })
+
+  it('Successfully fetches a parent chain with `getNetwork`', async function () {
+    const parentChain = await getNetwork(arbOneId)
+    expect(parentChain.chainID, fetchErrorMessage).to.be.eq(arbOneId)
+  })
+
+  it('Fails to fetch an Orbit chain with `getNetwork`', async function () {
```
fionnachan, line 77: i think the text has to be more descriptive to include that the Orbit chain is being fetched under the assumption that it's a layer 1

tests/unit/network.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        .to.be.undefined
+    }
+  })
+
+  it('Successfully fetches an Orbit chain with `getNetwork`', async function () {
+    const chain = await getNetwork(mockOrbitChainId)
+    expect(chain.chainID, fetchErrorMessage).to.be.eq(mockOrbitChainId)
+  })
+
+  it('Fails to fetch a parent chain with `getNetwork`', async function () {
```
fionnachan, line 98: similarly this should say a Base chain is being fetched under the assumption of it being a layer 2

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
+}
+const isChildChain = (chain: Chain): chain is ChildChain => {
+  return chain && 'partnerChainID' in chain
+}
+
+const isL1Chain = (chain: Chain): chain is L1Network => {
+  return chain && isParentChain(chain) && !isChildChain(chain)
+}
+
+const isL2Chain = (chain: Chain): chain is L2Network => {
```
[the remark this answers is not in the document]
brtkx: I wouldn't do that tbh, Orbit chains are Arbitrum chains too, can be confusing what that actually refers to

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
...
 
-const getNetwork = async (
+const isParentChain = (chain: Chain): chain is ParentChain => {
+  return chain && 'partnerChainIDs' in chain
+}
+const isChildChain = (chain: Chain): chain is ChildChain => {
+  return chain && 'partnerChainID' in chain
+}
+
+const isL1Chain = (chain: Chain): chain is L1Network => {
```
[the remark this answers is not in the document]
spsjvc: We can do that as part of the renaming effort, don't wanna mix up the terms yet

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,100 @@ export const l2Networks: L2Networks = {
   },
 }
 
-const getNetwork = async (
+const isParentChain = (chain: Chain): chain is ParentChain => {
+  return chain && 'partnerChainIDs' in chain
+}
+const isChildChain = (chain: Chain): chain is ChildChain => {
+  return chain && 'partnerChainID' in chain
```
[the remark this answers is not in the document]
fionnachan: yes, i mean in this case, what we want to test for is that this chain is a child chain of some chain, so the name should reflect that
just reading the function name makes me think that it is the child chain of a set pair of chains mentioned in a function whenever i am calling this function

src/lib/dataEntities/networks.ts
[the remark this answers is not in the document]
fionnachan: fair enough

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,89 @@ export const l2Networks: L2Networks = {
...
+  const children = Object.values(childChains).filter(
+    childChain => childChain.partnerChainID === chain.chainID
+  )
+  return children
+}
+
+export let l1Networks: L1Networks = getL1Chains()
+export let l2Networks: L2Networks = getL2Chains()
+
+export const getNetwork = async (
```
spsjvc, line 432: `getL1Network` should return a value properly if the passed in provider is a provider for an L2 that is a parent chain (e.g. Arbitrum Goerli)

src/lib/dataEntities/networks.ts
```diff
@@ -329,9 +349,87 @@ export const l2Networks: L2Networks = {
...
   signerOrProviderOrChainID: SignerOrProvider | number,
-  layer: 1 | 2
+  layer?: 1 | 2
```
spsjvc, line 432: We should keep this required but treat it as "parent chain" and "child chain".

If you pass in 1, then `getL1Network` should work for both an L1 chain and an L2 that is a parent chain for Orbit chains.
If you pass in 2, then `getL2Network` should work for either an L2 or an Orbit chain.

That makes it work with all existing code without any changes. I know it's confusing in the interim before everything is renamed.

src/lib/dataEntities/networks.ts
```diff
@@ -395,12 +501,46 @@ export const getEthBridgeInformation = async (
...
+ * @param customL2Network
+ */
 export const addCustomNetwork = ({
   customL1Network,
   customL2Network,
 }: {
   customL1Network?: L1Network
-  customL2Network: L2Network
+  customL2Network: L2Network | OrbitChain
 }): void => {
```
spsjvc, line 544: I think we may have to do something like this:
```
type AddCustomNetworkProps = {
  customL1Network?: L1Network
  customL2Network: L2Network
} | {
  customL1Network?: L2Network
  customL2Network: OrbitChain
}

export const addCustomNetwork = ({
  customL1Network,
  customL2Network,
}: AddCustomNetworkProps): void
```

spsjvc: So either you pass in an L1 and an L2 (e.g. Sepolia and Arbitrum Sepolia) or you pass in an L2 and an Orbit chain

Again the case of misnomer, we should think of `customL1Network` as `parentChain` and `customL2Network` as `childChain`.

tests/unit/network.test.ts
spsjvc, line 1: think of `getL1Network` as `getParentChain` and `getL2Network` as `getChildChain`

- getL1Network with an L1 network should work
- getL1Network with arb1 shouldn't work
- getL1Network with an L2 network that is a parent chain should work
- getL2Network should not work for an L1
- getL2Network should work for an L2
- getL2Network should work for an Orbit chain

spsjvc: We should also test for incorrect pairs (L1<>L3)

src/lib/dataEntities/networks.ts
```diff
@@ -507,6 +640,25 @@ export const addDefaultLocalNetwork = (): {
...
 ): network is L1Network => {
-  if ((network as L1Network).partnerChainIDs) return true
-  else return false
+  if (!network.partnerChainIDs) {
+    return false
+  }
+  return network.partnerChainIDs.length > 0
+}
+
+const createNetworkStateHandler = () => {
```
spsjvc, line 649: Move this to tests file

yahgwai (approved):

PR #381, opened 2023-12-01
feat: return status in L2ToL1Message.waitUntilReadyToExecute

DZGoldman:


src/lib/message/L2ToL1MessageClassic.ts
```diff
@@ -349,18 +349,20 @@ export class L2ToL1MessageReaderClassic extends L2ToL1MessageClassic {
    * WARNING: Outbox entries are only created when the corresponding node is confirmed. Which
    * can take 1 week+, so waiting here could be a very long operation.
    * @param retryDelay
-   * @returns
+   * @returns outbox entry status (either executed or confirmed but not executed)
```
yahgwai, line 352: "either executed or confirmed but not executed" - should be not pending?

yahgwai (approved):
Looks good to me, but prob best to get an approve from the full stack team for sdk changes

spsjvc (approved):

PR #385, opened 2023-12-18
update inbox utils to allow L2 as parent

spsjvc:


src/lib/inbox/inbox.ts
```diff
@@ -88,9 +103,11 @@ export class InboxTools {
     const diff = block.timestamp - blockTimestamp
     if (diff < 0) return block
 
+    const blockTime = this.l1NetworkIsArbitrum ? 12 : 0.25
```
gzeoneth: shouldn't this be in reverse? if it is Arbitrum, the block time can be assumed to be 0.25s
```suggestion
    const blockTime = this.l1NetworkIsArbitrum ? 0.25 : 12
```

gzeoneth (changes requested):

PR #393, opened 2024-01-16
Deploy both L1<->L2 and L2<->L3 token bridges in `testSetup`

godzillaba:
the weth withdrawal gas limit had to be raised because the weth contract has changed on the parent chain when testing orbit


godzillaba:
see #400

PR #404, opened 2024-01-25
feat: get l2 block range from node

brtkx:


scripts/genAbi.js
```diff
@@ -28,12 +28,12 @@ async function main() {
   // instead of writing in postinstall in each of those packages, we should target a local folder in sdk's postinstall
 
   console.log('building nitro')
-  execSync(`${npmExec} run hardhat:prod compile`, {
```
brtkx, line 31: [old command](https://github.com/OffchainLabs/nitro-contracts/blob/v1.0.0/package.json#L24) replaced with a [new one](https://github.com/OffchainLabs/nitro-contracts/blob/main/package.json#L21)

PR #445, opened 2024-04-03
feat(v3): non-18 decimal custom gas token support

brtkx:
Closes FS-406


tests/integration/eth.test.ts
```diff
@@ -45,6 +46,7 @@ dotenv.config()
 describe('Ether', async () => {
   beforeEach('skipIfMainnet', async function () {
     await skipIfMainnet(this)
+    await skipIfNon18Decimals(this)
```
brtkx, line 49: Skip eth tests here

tests/integration/customerc20.test.ts
```diff
@@ -53,6 +54,7 @@ const withdrawalAmount = BigNumber.from(10)
 describe('Custom ERC20', () => {
   beforeEach('skipIfMainnet', async function () {
     await skipIfMainnet(this)
+    await skipIfNon18Decimals(this)
```
brtkx, line 57: Not sure about this one. `registerCustomToken` doesn't seem to support non 18 decimals but it's got nothing to do with the native currency.

Should we make `registerCustomToken` work with non-18 decimals with this PR, or leave as is?

brtkx: Also it's skipped because we're passing tokens based on native currency in tests so it's failing decimals.

brtkx: **TBC: I need to go through it again to confirm the above**

src/lib/utils/lib.ts
```diff
@@ -195,3 +196,51 @@ export const getBlockRangesForL1Block = async (
...
+  if (!nativeTokenAddress || nativeTokenAddress === constants.AddressZero) {
+    return 18
+  }
+
+  const nativeTokenContract = ERC20__factory.connect(
+    nativeTokenAddress,
+    l1Provider
+  )
+
+  return nativeTokenContract.decimals()
```
gvladika, line 218: ERC20 don't need to necessary implement `decimals()`. If it's not implemented (so call reverts) we take that as token using 0 decimals.

src/lib/utils/lib.ts
```diff
@@ -195,3 +196,51 @@ export const getBlockRangesForL1Block = async (
...
+    const scaledAmount = amount.div(multiplier)
+    // round up if necessary
+    if (scaledAmount.mul(multiplier).lt(amount)) {
+      return scaledAmount.add(BigNumber.from(1))
+    }
+    return scaledAmount
+  }
+
+  // decimals > 18
+  return amount.mul(multiplier)
```
gvladika, line 245: If decimals > 18 then we scale up by 10^(decimals-18)

tests/integration/custom-fee-token/customFeeTokenTestHelpers.ts
```diff
@@ -14,6 +14,8 @@ const ethProvider = () => new StaticJsonRpcProvider(config.ethUrl)
 const arbProvider = () => new StaticJsonRpcProvider(config.arbUrl)
 const localNetworks = () => getLocalNetworksFromFile()
 
+const DECIMALS = Number(process.env.DECIMALS)
```
gvladika, line 17: You can also read decimals from contract

tests/integration/customerc20.test.ts
gvladika: Registering custom token is sending retryable ticket to child chain. Tickets needs to be paid for in fee token. Amount of fee token needs to be calculated (and potentially scaled) and provided to the smart contract call. Here's one example: https://github.com/OffchainLabs/token-bridge-contracts/blob/657972fca20cea08b8ac118e6344af0584a3b5b6/test-e2e/orbitTokenBridge.ts#L350

gvladika (commented):
Adding checks for fee token would be useful, ie. to confirm that the proper (scaled) amount of fee tokens has been spent by user when ie. bridging tokens

Also, would be good to execute tests for >18 decimals (not necessary in CI if too time consuming, but at least once manually before releasing the feature).

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -1051,8 +1070,18 @@ export class AdminErc20Bridger extends Erc20Bridger {
...
+          decimals: nativeTokenDecimals,
+        }),
+        scaleToNativeDecimals({
+          amount: setGatewayGas.gasLimit
+            .mul(doubleFeePerGas)
+            .add(setGatewayGas.maxSubmissionCost),
+          decimals: nativeTokenDecimals,
+        }),
         l1SenderAddress,
       ])
```
brtkx, line 1087: Do we need to set `value` in the return here? I think it only should be set if L3 chains use ETH as the native currency?

tests/integration/customerc20.test.ts
```diff
@@ -71,10 +68,6 @@ describe('Custom ERC20', () => {
     }
     await fundL1(testState.l1Signer)
     await fundL2(testState.l2Signer)
-
-    if (isL2NetworkWithCustomFeeToken()) {
```
brtkx, line 75: We already do that in `testSetup`

tests/integration/retryableData.test.ts
```diff
@@ -34,6 +34,7 @@ import { isL2NetworkWithCustomFeeToken } from './custom-fee-token/customFeeToken
 describe('RevertData', () => {
   beforeEach('skipIfMainnet', async function () {
     await skipIfMainnet(this)
+    await skipIfNon18Decimals(this)
```
brtkx, line 37: This test doesn't work for non 18 decimals. The data we check against is hardcoded and doesn't match.

Safe to omit and only check for 18 decimal chains?

gvladika (dismissed):
lgtm

tests/integration/retryableData.test.ts
```diff
@@ -34,6 +34,7 @@ import { isL2NetworkWithCustomFeeToken } from './custom-fee-token/customFeeToken
 describe('RevertData', () => {
   beforeEach('skipIfMainnet', async function () {
     await skipIfMainnet(this)
+    await skipIfCustomGasToken(this)
```
spsjvc, line 37: seems dangerous to skip this

tests/integration/eth.test.ts
```diff
@@ -99,15 +103,21 @@ describe('Ether', async () => {
...
 
     const initialInboxBalance = await l1Signer.provider!.getBalance(
       inboxAddress
     )
-    const ethToDeposit = parseEther('0.0002')
+    const amount = '0.0002'
+    const ethToDeposit = scaleToNativeTokenDecimals({
+      amount: parseEther(amount),
+      decimals,
+    })
```
gvladika, line 120: Don't need to do scaling here. If ETH is used, 'parseEther' is used so `ethToDeposit` will be equal 0.0002 * 10^18 wei.  If fee token is used you can multiply amount by 10 ** decimals. Ie. if you want to deposit 0.0002 USDC `ethToDeposit` would be 0.0002 * 10 ** 6 wei. That will result with 0.0002 * 10 ** 18 wei minted on the L2.

tests/integration/eth.test.ts
```diff
@@ -155,7 +171,11 @@ describe('Ether', async () => {
...
```
gvladika, line 178: Same here

src/lib/assetBridger/erc20Bridger.ts
brtkx: Updated so that we only pass value when ETH is the native currency

tests/integration/l1l3Bridger.test.ts
```diff
@@ -356,7 +366,16 @@ describe('L1 to L3 Bridging', () => {
...
       'should properly get l2 and l1 fee token addresses',
-      async () => {
+      async function () {
+        const decimals = await getNativeTokenDecimals({
+          l1Provider: l1Signer.provider!,
+          l2Network: l3Network,
+        })
+
+        if (decimals !== 18) {
+          this.skip()
```
brtkx, line 376: For non-18 decimals these tests throw: `Error: L1 gas token has incorrect decimals. Use skipGasToken when depositing`.

We skip all the tests that throw this error for non-18 decimals.

gvladika (approved):
LGTM

PR #446, opened 2024-04-04
feat: add L3 networks as L1 partner chains to enable teleportation

dewanshparashar:
Till now, chain pairs were only restricted to L1-L2 or L2-L3, but now L1-L3 is also possible.


godzillaba (commented):
lgtm, but this changes the definition of `partnerChainIDs`. would adding a new optional property like `indirectChildrenChainIDs[]` or something like that work?

PR #460, opened 2024-04-19
feat: get arbitrum network information from rollup

spsjvc:
This PR renames `getEthBridgeInformation` to `getArbitrumNetworkInformationFromRollup` and augments it by returning both the `parentChainId` and the `confirmPeriodBlocks`. I think it could be really useful in situations like [this one](https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/a0133c44dee83401096da2c8c84d020cb209f8fb/src/utils/registerNewNetwork.ts#L6).

The tests were moved to another file and the necessary assertions were added.


TucksonDev (approved):
LGTM

PR #474, opened 2024-05-23
fix: inconsistent `erc20L1Address` casing

gomesalexandre:
Does what it says on the box - fixes inconsistent casing of `erc20L1Address` between `Erc20DepositParams` and `Erc20WithdrawParams`


hkalodner:
@cla-bot check

spsjvc:
Hi @gomesalexandre, ty for your contribution! Doing this would be a breaking change, so it's gonna be fixed in the next major version which we're currently working on.

PR #475, opened 2024-05-24
 fix: adjust `blockTime` inside `InboxTools.findFirstBlockBelow` in v3

brtkx:


src/lib/inbox/inbox.ts
```diff
@@ -154,11 +191,25 @@ export class InboxTools {
...
     )
 
+    const isParentChainArbitrum = await isArbitrumChain(this.l1Provider)
+
+    if (isParentChainArbitrum) {
+      const arbProvider = new ArbitrumProvider(
+        this.l1Provider as JsonRpcProvider
+      )
+      const currentArbBlock = await arbProvider.getBlock(
+        await arbProvider.getBlockNumber()
```
gzeoneth, line 208: I think we can use "latest" instead of making 2 calls?

src/lib/inbox/inbox.ts
```diff
@@ -183,7 +234,8 @@ export class InboxTools {
       await multicall.multiCall(multicallInput, true)
 
     const firstEligibleBlockNumber =
-      currentBlockNumber.toNumber() - maxTimeVariation.delayBlocks.toNumber()
+      (currentL1BlockNumber ?? currentBlockNumber.toNumber()) -
```
gzeoneth, line 237: use `isParentChainArbitrum ?` instead of `currentL1BlockNumber ??`, doesn't matter too much tho

src/lib/inbox/inbox.ts
```diff
@@ -90,6 +95,38 @@ export class InboxTools {
     blockNumber: number,
     blockTimestamp: number
   ): Promise<Block> {
+    const isParentChainArbitrum = await isArbitrumChain(this.l1Provider)
+
+    const nodeInterface = NodeInterface__factory.connect(
```
gzeoneth, line 100: can be defined within the next `if`

src/lib/inbox/inbox.ts
gzeoneth, line 1:
```
const diffBlocks = Math.max(Math.ceil(diff / this.l1Network.blockTime), 10)
```
I think we need to hardcode the blocktime here because we want to real "L1" block time here but `l1Network.blockTime` for a L3 chain can be the L2 block time?

spsjvc (approved):

PR #481, opened 2024-06-12
feat: check if token is registered on gateways

brtkx:


src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -840,6 +841,40 @@ export class Erc20Bridger extends AssetBridger<
...
+      erc20L1Address,
+      l2Provider
+    )
+
+    const l2AddressFromL2Gateway = await L2ERC20Gateway__factory.connect(
+      l2GatewayAddressFromL2Router,
+      l2Provider
+    ).calculateL2TokenAddress(erc20L1Address)
+
+    return tokenL2AddressFromL1GatewayRouter === l2AddressFromL2Gateway
```
spsjvc, line 876: we should lowercase both just in case

src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -840,6 +841,40 @@ export class Erc20Bridger extends AssetBridger<
...
   }
+
+  /**
+   * Checks if custom gateway has been registered
+   * @param erc20L1Address
+   * @param l1Provider
+   * @param l2Provider
+   * @returns
+   */
+  public async isCustomGatewayRegistered({
```
spsjvc, line 852:
```suggestion
  /**
   * Checks if the token has been properly registered on both gateways. Mostly useful for tokens that use a custom gateway.
   *
   * @param erc20L1Address
   * @param l1Provider
   * @param l2Provider
   * @returns
   */
  public async isRegistered({
```

spsjvc (approved):
LGTM

PR #485, opened 2024-06-13
feat: renames in `isRegistered` method

brtkx:
closes FS-605


src/lib/assetBridger/erc20Bridger.ts
```diff
@@ -767,6 +802,18 @@ export class Erc20Bridger extends AssetBridger<
       params.parentSigner
     )
 
+    if (
+      !this.isRegistered({
```
spsjvc, line 806: this needs to be awaited

spsjvc (approved):

PR #493, opened 2024-06-18
fix: bump new advisory ws to patched version

chrstph-dvx:
- Bump ws versions
- Add GHSA-3787-6prv-h9w3 to ignored audit error


audit-ci.jsonc
fionnachan, line 1: would be good to review the whole list and see if we can bump them up to the patched version and remove them from this list

fionnachan (approved):

PR #502, opened 2024-07-02
feat: enable teleporter to work with v4

douglance:


scripts/testSetup.ts
spsjvc, line 1: I think we shouldn't change this file. We can just hardcode the necessary local L1 info in the L1-L3 bridger test suite

spsjvc (approved):

PR #526, opened 2024-08-02
Create `assembleChildTx` to abstract away signing logic in an attempt to better support browser wallets

matzapata:
Extracted the assembling logic for `signingChildTx` away into it's separate function to better support browser wallets. 

Metamask and other wallets currently don't support signing without sending ([issue](https://github.com/MetaMask/metamask-extension/issues/3475#issuecomment-372051711)) and so `signChildTx` is impossible without users exporting private keys. 

This Pr basically renames `signChildTx` to `assembleChildTx` removing the final line so the decision of only signing or signing and sending moves back to the developer integrating the SDK. I understand just signing is what makes the most sense but I believe sending doesn't make any harm and allows these extra cases.

### DIFF
--- a/src/lib/inbox/inbox.ts
+++ b/src/lib/inbox/inbox.ts
@@ -415,9 +415,8 @@ export class InboxTools {
   }
 
   /**
-   * Sign a transaction with msg.to, msg.value and msg.data.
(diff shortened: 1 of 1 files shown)


Flisflas69 (commented):
Ok

Flisflas69 (commented):
Ok

PR #530, opened 2024-08-28
feat: non-18 decimal custom gas token support 

spsjvc:


tests/integration/customerc20.test.ts
```diff
@@ -75,10 +72,6 @@ describe('Custom ERC20', () => {
     }
     await fundParentSigner(testState.parentSigner)
     await fundChildSigner(testState.childSigner)
-
-    if (isArbitrumNetworkWithCustomFeeToken()) {
```
brtkx, line 79: We don't need this, `fundParentSigner` and `fundChildSigner` cover this

brtkx (approved):
Approving on behalf of @spsjvc

PR #552, opened 2024-11-14
initial commit for viem api

chrstph-dvx:


src/experimental/arbitrumDeposit/depositEth.ts
chrstph-dvx, line 1: Extracting action in their own function, means user can either call it with `depositEth(client, params)` or `client.depositEth(params)`

PR #554, opened 2024-12-10
refactor: set up monorepo

douglance:
- minimal change in order to turn the repository into a monorepo. 
- we migrate the project at the root into `packages/sdk` and move its unique config there
- we'll split out the `scripts` folder in a future PR


yarn.lock
```diff
@@ -1,7 +1,6 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # yarn lockfile v1
 
-
```
spsjvc, line 4: hm how did this end up here? can you revert it

.github/workflows/build-test.yml
```diff
@@ -25,6 +25,8 @@ jobs:
 
       - name: Install node_modules
         uses: OffchainLabs/actions/node-modules/install@main
+        with:
+          cache-key: ${{ github.run_id }}-${{ github.run_number }}
```
spsjvc:
```suggestion
          cache-key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}-monorepo
```

package.json
```diff
@@ -87,9 +50,6 @@
     "typescript": "^4.9.5",
     "yargs": "^17.3.1"
   },
-  "files": [
-    "dist/**/*"
-  ],
   "resolutions": {
```
spsjvc, line 53: should we move resolutions for runtime dependencies to the package's package.json?

packages/sdk/src/lib/utils/env.ts
```diff
@@ -0,0 +1,6 @@
+import * as dotenv from 'dotenv'
+import * as path from 'path'
+
+export const loadEnv = () => {
+  dotenv.config({ path: path.resolve(__dirname, '../../../../../.env') })
```
spsjvc, line 5: will this work properly both in the case when you're running commands at the root level of the monorepo, and when you cd into the package and run it there?

spsjvc (approved):

PR #558, opened 2024-12-18
feat: adds viem compatibility package

douglance:


packages/viem-compatibility/package.json
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  "scripts": {
+    "build": "rm -rf dist && tsc -p tsconfig.json",
+    "test:unit": "mocha -r ts-node/register 'tests/**/*.test.ts'",
+    "lint": "eslint .",
+    "format": "prettier './**/*.{js,json,md,ts,yml}' '!./src/lib/abi' --write && yarn run lint --fix"
+  },
+  "dependencies": {
+    "viem": "^2.21.55"
+  }
```
spsjvc, line 33: shouldn't this have both viem and ethers as peer dependencies?

packages/viem-compatibility/src/compatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+  Chain,
+  Client,
+  PublicClient,
+  Transport,
+  Log as ViemLog,
+  TransactionReceipt as ViemTransactionReceipt,
+} from 'viem'
+
+// based on https://wagmi.sh/react/ethers-adapters#reference-implementation
+export function publicClientToProvider<TChain extends Chain | undefined>(
```
spsjvc, line 17: the orbit sdk added some extra stuff here: https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/ethers-compat/publicClientToProvider.ts

and there's also a test that we could copy over https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/ethers-compat/publicClientToProvider.unit.test.ts

packages/viem-compatibility/src/compatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+    object.transport !== null &&
+    typeof object.transport === 'object' &&
+    'url' in object.transport &&
+    typeof object.transport.url === 'string' &&
+    object.type === 'publicClient'
+  )
+}
+
+export const transformPublicClientToProvider = (
+  provider: PublicClient | Client
```
spsjvc, line 50: not sure if we need this function at all

if we change the argument type to just `PublicClient` (which I think we should) then it's not needed

spsjvc (approved):

PR #565, opened 2024-12-26
feat: initialize viem package

douglance:


packages/sdk-viem/src/createArbitrumClient.ts
```diff
@@ -0,0 +1,59 @@
...
+  childWalletClient?: WalletClient
+}
+
+export function createArbitrumClient({
+  parentChain,
+  childChain,
+  parentRpcUrl,
+  childRpcUrl,
+  parentWalletClient,
+  childWalletClient,
```
chrstph-dvx, line 37: I don't think we need the childWalletClient here

chrstph-dvx: Actually, I would not request any client here, and just create everything within the function

packages/sdk-viem/src/createArbitrumClient.ts
```diff
@@ -0,0 +1,59 @@
...
+}: CreateArbitrumClientParams): ArbitrumClients {
+  const parentPublicClient = createPublicClient({
+    chain: parentChain,
+    transport: http(parentRpcUrl || parentChain.rpcUrls.default.http[0]),
+  })
+
+  const childPublicClient = createPublicClient({
+    chain: childChain,
+    transport: http(childRpcUrl || childChain.rpcUrls.default.http[0]),
+  }).extend(arbitrumParentClientActions())
```
chrstph-dvx, line 47: We extend the child client with arbitrum**Parent** actions?

packages/sdk-viem/src/actions.ts
```diff
@@ -0,0 +1,223 @@
...
+  sendCrossChainTransaction: (
+    params: SendCrossChainTransactionParameters
+  ) => Promise<CrossChainTransactionStatus>
+
+  depositEth: (
+    params: DepositEthParameters
+  ) => Promise<CrossChainTransactionStatus>
+}
+
+async function prepareDepositEthTransaction(
```
chrstph-dvx, line 75: I would still reexport the actions on their own. It helps with treeshaking and it's common viem API to have both decorators and actions, `https://viem.sh/docs/clients/custom#tree-shaking`

packages/sdk-viem/src/createArbitrumClient.ts
```diff
@@ -6,15 +6,13 @@ import {
...
-  ArbitrumDepositActions,
   ArbitrumParentWalletActions,
-  arbitrumParentClientActions,
   arbitrumParentWalletActions,
 } from './actions'
 
 export type ArbitrumClients = {
   parentPublicClient: PublicClient
-  childPublicClient: PublicClient & ArbitrumDepositActions
+  childPublicClient: PublicClient
```
chrstph-dvx, line 15: We don't need the `childPublicClient` here, we recreate it on L42

packages/sdk-viem/src/createArbitrumClient.ts
chrstph-dvx: If they want to keep control on their client, they can extend with actions on their own. I would expect `createArbitrumClient` to do every thing for me and create the clients not extend them. But it's not an issue.

PR #582, opened 2025-03-16
fix sdk doc link

cocoyeal:
this original link goes 404, it should be updated :)


cocoyeal:
I have signed all the CLAs for the individuals above :)

fionnachan:
it no longer goes to 404 as we have fixed the link, thanks!

PR #583, opened 2025-03-31
fix: properly handle bold genesis assertion

spsjvc:


gzeoneth (approved):
LGTM, some comment to explain can be nice

PR #607, opened 2025-07-30
remove legacy patchwork

godzillaba:


packages/sdk/scripts/genNetwork.ts
```diff
@@ -35,66 +27,15 @@ function getLocalNetworksFromContainer(which: 'l1l2' | 'l2l3'): any {
...
-  patchedL3Network?: ArbitrumNetwork
-}> {
-  const patchedL2Network = mapL2NetworkToArbitrumNetwork(l2Network)
-
-  // native token for l3
-  if (l3Network && l2Provider) {
-    const patchedL3Network = mapL2NetworkToArbitrumNetwork(l3Network)
-
-    try {
-      patchedL3Network.nativeToken = await IERC20Bridge__factory.connect(
```
TucksonDev, line 58: We need this to get the nativeToken from the local chain (if it exists), so the tests use the appropriate contracts.

PR #623, opened 2026-02-13
feat: add reverse tracing for bridge transactions

douglance:
Adds methods to trace bridge transactions back to their originating chain:

- `ChildTransactionReceipt.getParentTransactionHash` – traces retryable redeems back to the parent tx
- `ChildTransactionReceipt.getParentDepositTransactionHash` – traces ETH deposits (type 0x64) back to the parent tx
- `ParentTransactionReceipt.getChildWithdrawTransactionHash` – traces withdrawal executions back to the child tx

Works for both standard ETH and custom gas token chains (including non-18-decimal tokens).


TucksonDev (approved):
LGTM 🚀

